PR #1, opened 2018-10-11
Improve javascript/node example

soyuka:


examples/nodejs/publisher.js
```diff
@@ -2,17 +2,16 @@ const http = require('http');
 const querystring = require('querystring');
 
 const postData = querystring.stringify({
-    'topic': 'http://localhost:3000/demo/books/1.jsonld',
+    'topic': 'https://example.com/books/1.jsonld',
```
dunglas: Can you revert this change? Or use an env var Because currently it works "out of the box" with the demo in dev mode.

examples/nodejs/publisher.js
```diff
@@ -2,17 +2,16 @@ const http = require('http');
...
 });
 
 const req = http.request({
     hostname: 'localhost',
-    port: '3000',
     path: '/publish',
     method: 'POST',
     headers: {
-        Authorization: 'Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.HB0k08BaV8KlLZ3EafCRlTDGbkd9qdznCzJQ_l8ELTU',
+      Authorization: 'Bearer ' + process.env.PUBLISHER_JWT_TOKEN,
```
dunglas: Can you use the existing token as default value? It allows to test easily.

examples/nodejs/README.md
```diff
@@ -0,0 +1,35 @@
+# Mercure with nodejs
```
dunglas: Mercure with Node.js

examples/nodejs/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+First thing is to run the hub, the easiest way is to use docker:
+
+```bash
+docker run \
+    -e PUBLISHER_JWT_KEY=jwtkey -e SUBSCRIBER_JWT_KEY=jwtkey -e ALLOW_ANONYMOUS=1 -e CORS_ALLOWED_ORIGINS="*" \
+        -p 80:80 -p 443:443 \
```
dunglas: Indent problem?

examples/nodejs/index.html
```diff
@@ -0,0 +1,7 @@
+<script type="text/javascript">
```
dunglas: `type` can be removed.

examples/nodejs/index.html
```diff
@@ -0,0 +1,7 @@
...
+const url = new URL('http://localhost/subscribe');
+url.searchParams.append('topic', 'https://example.com/books/{id}');
+
+const eventSource = new EventSource(url);
```
dunglas: Can you use the discovery mechanism? (available in the readme)

soyuka: I won't be able to run the fetch `example.com/books` though :|. What I had in mind was a simple example that you can run on your device to see the magic happened.

dunglas: Yes It's why I used localhost in the first time. I think we should replace `example.com` by `localhost` everywhere.

dunglas:
Hey @soyuka! Do you plan to finish this or should I close?

PR #8, opened 2018-10-15
Fix typo

teohhanhui:
SHOULD advertises -> SHOULD advertise


dunglas:
Thanks @teohhanhui, good catch!

PR #11, opened 2018-10-18
Support duplex communication

dunglas:
This PRs add proper duplex support to Mercure. This change will allow to easily dispatch updates from clients (dispatching updates from servers is still supported, of course).

It gives more flexibility to clients, and will help in some case where relatively low latency is necessary (such as video games).

The main changes are:

* Two authorization mechanisms are now available, both to publish and to subscribe: using an `Authorization` header (for servers), or using a cookie (for web browsers)
* The same JWT **may** be used to subscribe and to publish
* The format of the Mercure's JWT claim now looks like:
```json
{
    "mercure": {
        "publish": ["an empty array for public updates only, or a list of targets, or * for all targets"],
        "subscribe": ["a list of targets, or * for all targets (not required for public messages)"]
    },
    "other JWT claims": "..."
}
```
* The same URL must be used to subscribe (`GET`) and to publish (`POST`) (it was not mandatory before)
* Consequently, CSRF mitigation must be added to the hub, a new "Security Considerations" section has been added

I'll update the reference implementation in the same PR.

This change is the result of a discussion with @K-mos.


sroze:
Very good idea Marion!

PR #18, opened 2018-10-30
Fix CORS handling

### LINKED ISSUE #17: How to authenticate subscriber?
Hi,
could you please clarify how subscribers can authenticate themselves against hub in case the hub runs on a different domain?

I have a frontend app on one domain with REST API on another domain and a mercure hub on a third domain. At https://demo.mercure.rocks/ you have two options - mercureAuthorization using a cookie, or HTTP header. In my case I cannot send HTTP header to authenticate with EventSource (even you alert in your demo that it is not possible). But how could I make the hub to set the mercureAuthorization cookie for me? The discover endpoint may be on the API, but it is on a different domain.

Maybe I am doing something wrong, thanks for any help!

dunglas:
Fix #17

### DIFF
--- a/hub/server.go
+++ b/hub/server.go
@@ -91,10 +91,11 @@ func (h *Hub) chainHandlers() http.Handler {
 
 	var corsHandler http.Handler
 	if len(h.options.CorsAllowedOrigins) > 0 {
+		log.Print("cors")
 		allowedOrigins := handlers.AllowedOrigins(h.options.CorsAllowedOrigins)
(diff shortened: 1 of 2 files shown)


hub/server.go
```diff
@@ -91,10 +91,11 @@ func (h *Hub) chainHandlers() http.Handler {
 
 	var corsHandler http.Handler
 	if len(h.options.CorsAllowedOrigins) > 0 {
+		log.Print("cors")
```
soyuka: you're keeping this?

dunglas: oops

PR #24, opened 2018-11-08
Change "PWA" to "webapp" in the use cases

nhoizey:
It looks like all of this would be useful on any webapp, even if not Progressive.


dunglas:
Indeed, thanks!

nhoizey:
You're welcome!

Mercure looks awesome, we definitely have to try it.

PR #32, opened 2018-11-26
Add a chat demo using Flask (Python)

### LINKED ISSUE #12: Could you please provide a chat demo ?

dunglas:
This PR adds an example of a real-time chat app in 23 lines of Python (2 dependencies:  [the Flask microframework](http://flask.pocoo.org/) and PyJWT) and 35 lines of JavaScript (0 dependencies). Of course it benefits from built-in capabilities of the Mercure protocol: HTTP/2/HTTP/3 compatibility, automatic reconnection with re-fetching of missed messages, authentication...

[live demo](https://python-chat.mercure.rocks/)

mercure-chat-python

Flask is only used to handle the authentication part (cookie method) and to serve the HTML layout. The chat system itself uses directly the Mercure hub.

😻

Closes #12.


examples/chat-python-flask/templates/chat.html
```diff
@@ -0,0 +1,31 @@
...
+        No messages
+    </div>
+
+    <script type="application/json" id="config">
+        {{ config|tojson|safe }}
+    </script>
+    <script src="/static/chat.js"></script>
+</body>
+
+</html>
```
toofff: Add new line at the end fille please

examples/chat-python-flask/templates/chat.html
```diff
@@ -0,0 +1,31 @@
...
+
+    <form>
+        <input name="username" placeholder="username" size="10">
+        <input name="message" placeholder="Your message..." size="50" autocomplete="off">
+        <input type="submit" value="Post">
+    </form>
+    <hr>
+
+    <div id="messages">
+        No messages
```
Gregcop1: No message

examples/chat-python-flask/static/chat.js
```diff
@@ -0,0 +1,35 @@
...
+let ul = null
+es.onmessage = ({ data }) => {
+    const { username, message } = JSON.parse(data)
+    if (!username || !message) throw new Error('Invalid payload')
+
+    if (!ul) {
+        ul = document.createElement('ul')
+
+        const messages = document.getElementById('messages')
+        messages.firstChild.remove()
```
Gregcop1: `messages.innertHTML = '';` should be faster

examples/chat-python-flask/static/chat.js
```diff
@@ -0,0 +1,35 @@
...
+    const li = document.createElement('li')
+    li.append(document.createTextNode(`<${username}> ${message}`))
+    ul.append(li)
+}
+
+document.querySelector('form').onsubmit = function (e) {
+    e.preventDefault()
+
+    const body = new URLSearchParams({
+        data: JSON.stringify({ username: this.elements.username.value, message: this.elements.message.value }),
```
tlenclos: Why do you need to send the username if the user is logged ?

dunglas: It is authorized to subscribe and to post to the, not really "logged in" (I mean it is authorized but not authenticated).
A proper login system should be added to the Python app (if appropriate), then the username could be set in the dedicated JWT claim, and sent along with the message. But then I think that it would be more efficient (and more secure) to send new messages to the Python app, that will then send it (including the current username) to the hub.

examples/chat-python-flask/chat.py
```diff
@@ -0,0 +1,50 @@
...
+def chat():
+    targets = [os.environ.get('TARGET', 'chan')]
+    token = jwt.encode(
+        {'mercure': {'subscribe': targets, 'publish': targets}},
+        os.environ.get('JWT_KEY', '!UnsecureChangeMe!'),
+        algorithm='HS256'
+    )
+
+    hub_url = os.environ.get('HUB_URL', 'http://localhost:3000/hub')
+    topic = os.environ.get('HUB_URL', 'http://example.com/chat')
```
aslafy-z: I think this env variable has to be TOPIC, not HUB_ URL.

PR #39, opened 2018-12-27
Second pass of specification doc

superhaggis:
As promised :)

You'll see by my second commit that I reverted my "JWT to JWS" change - JWT hadn't been mentioned previously in the document and thus was undefined and may have been referenced in error.  If JWT is correct, then "JSON Web Token (JWT)" should be used in the first instance with a link to https://tools.ietf.org/html/rfc7519.

Let me know if you'd like this change to be made.

### DIFF
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -46,7 +46,7 @@ The keywords **MUST**, **MUST NOT**, **REQUIRED**, **SHALL**, **SHALL NOT**, **S
 ![Discovery Schema](discovery.png)
 
 If the publisher is a server, it **SHOULD** advertise the URL of one or more hubs to the subscriber, allowing it to receive live updates when topics are updated.
-If more than one hub URL is specified, it is expected that the publisher notifies each hub, so the subscriber **MAY** subscribe to one or more of them.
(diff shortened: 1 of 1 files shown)


spec/mercure.md
```diff
@@ -46,7 +46,7 @@ The keywords **MUST**, **MUST NOT**, **REQUIRED**, **SHALL**, **SHALL NOT**, **S
 ![Discovery Schema](discovery.png)
 
 If the publisher is a server, it **SHOULD** advertise the URL of one or more hubs to the subscriber, allowing it to receive live updates when topics are updated.
-If more than one hub URL is specified, it is expected that the publisher will notify each hub, so the subscriber **MAY** subscribe to one or more of them.
```
dunglas: This sentence also comes from the WebSub spec:

> The URL of one or more hubs designated by the publisher. If more than one hub URL is specified, it is expected that the publisher notifies each hub, so the subscriber may subscribe to one or more of them.

May can we keep the wording from the W3C?

dunglas:
Thank you very much @superhaggis!

PR #53, opened 2019-03-05
Add a heartbeat mechanism, and timeouts support

dunglas:
Some reverse proxies kill connections after a (short) period of inactivity. Also, sometimes it's necessary to kill connections after some time to cleanup the memory.

This PR introduces new environment variables to start a heartbeat mechanism (that will prevent reverse proxies and some old browsers to kill the connection), and as well as read and write timeouts.

They are all disabled by default.

* `HEARTBEAT_INTERVAL`: interval between heartbeats (useful with some proxies, and old browsers), set to `0s` to disable (default), example `15s`
* `READ_TIMEOUT`: maximum duration for reading the entire request, including the body, set to `0` to disable (default), example: `2m`
* `WRITE_TIMEOUT`: maximum duration before timing out writes of the response, set to `0` to disable (default), example: `2m`


karser:
@dunglas what do you think of setting `HEARTBEAT_INTERVAL=30s` by default? Heartbeat is required for both [native](https://stackoverflow.com/questions/8761025/why-is-eventsource-connection-closed-every-30-60-sec-when-no-data-transported-w) and [polyfill](https://github.com/Yaffle/EventSource/issues/76) event source implementations.

dunglas:
Sound good to me!

PR #55, opened 2019-03-08
Fix IE browser support

rvanlaak:
Browser support mentioned seems wrong. Maybe refer to the link below?

> Browser support:
>> IE 10+, Firefox 3.5+, Chrome 3+, Safari 4+, Opera 12+
>> IE 8 - IE 9: XDomainRequest is used internally, which has some limitations (2KB padding is requried, no way to send cookies, no way to use client certificates)

https://github.com/Yaffle/EventSource#browser-support

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ In addition, a managed and high-scalability version of Mercure is [available in 
 * designed with [hypermedia in mind](https://en.wikipedia.org/wiki/HATEOAS), also supports [GraphQL](https://graphql.org/)
 * auto-discoverable through [web linking](https://tools.ietf.org/html/rfc5988)
 * message encryption support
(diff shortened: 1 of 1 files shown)


dunglas:
Yaffle's polyfill doesn't support IE7. But this one does: https://github.com/remy/polyfills/blob/master/EventSource.js

rvanlaak:
Ok great, thanks!

I see there is no link to a suggested polyfill. Would you like me to add these two options to the readme? Or don't you want to suggest a preferable polyfill?

dunglas:
The current one is more popular. Maybe could we add a note somewhere telling that this remy's polyfill supports IE7.

dunglas:
Closing for now, feel free to open a new PR to add the mention.

PR #61, opened 2019-03-16
fix typo CORS_ALLOWED_ORIGINS environment variable

damosse31:
CORS_ALLOWED_ORIGINS instead of CORS_ALLOWED_ORIGIN in README


dunglas:
Good catch, thanks!

PR #73, opened 2019-04-03
Mercure running via supervisor on linux

cklm:
added an example how to run mercure via supervisor on linux


cklm:
No problem - is added to readme

dunglas:
Thanks @cklm!

PR #89, opened 2019-05-16
Add missing allowed header

corebreaker:
Hello, i put Cache-Control in allowed headers to solve #90.


dunglas:
Thanks @corebreaker, good catch!

Issue #52, opened 2019-03-05
Immutable infrastructure; how to host mercure on them ?


tristanbes:
Hello,

I just asked a cloud provider if they have any plans to offer Mercure on their PaaS. They said that they cannot because they have an immutable architecture, each time we deploy, it's on a new container so data are not kept.

Reference tweet: https://twitter.com/Soulou/status/1102933346215231488

They pointed out issues discussed in #10, but we're not interested in redundancy, just host mercure on an immutable infrastructure :)

Is this possible ? Thanks

dunglas:
Hi,

Indeed the current boltdb requires write access to the disk. However this behavior would be easy to change. (The database is only use as a temporary history, it's not true persistent data).
If the disk is read only, and it's not possible to use a tmpfs or something similar (it would be the best solution), it's still possible to:

* implement a in-memory history (here is the interface, providing such implementation would be very easy: https://github.com/dunglas/mercure/blob/master/hub/history.go#L13)
* disable the history (the implementation is already provided: https://github.com/dunglas/mercure/blob/master/hub/history.go#L24, but not exposed as a configuration option for now)

A patch is very welcome, or I'll try to add this when I'll have a bit more free time. Thanks for reporting!

tristanbes:
I can't speak for them, but I think the disk is not read only, It's "just" lost on each deployments.

So basically you're saying we don't care loosing this file between each mercure hub deployments since it's only temporary history ? It doens't compromise the data ?

dunglas:
Exactly. It's only used for by the client after a network issue, when it tries to fetch lost updates: https://github.com/dunglas/mercure/blob/master/spec/mercure.md#reconnection-and-state-reconciliation

tristanbes:
Thanks for the clarification ! 🎉

Soulou:
Perfect, thanks @tristanbes  and @dunglas  for those specifications.

PR #92, opened 2019-05-22
Make it deployable on PaaS like Scalingo

Soulou:
As a follow up of what has been done https://github.com/dunglas/mercure/issues/52

This PR is doing two things:

- Adding compatibility with the PaaS Scalingo (with the `Procfile` file)
- Add a one-click deployment button in the README next to Docker/Kubernetes deployment methods

This could help to make Mercure more famous and easier to deploy and try.

If you want to try how the button is working (as it requires the scalingo.json to be present in the repo), here is a working version in our fork: https://github.com/Scalingo/mercure-scalingo

If you've any question, don't hesitate.

Note: Thanks to the `Procfile`, it would also make the application deployable on `Heroku`

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: ADDR=":${PORT}" mercure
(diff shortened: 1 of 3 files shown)


dunglas:
See #226. It would better fit in the documentation.

PR #110, opened 2019-06-27
Update README add Laravel Broadcaster to tools section 

mvanduijker:
As suggested in https://github.com/mvanduijker/laravel-mercure-broadcaster/issues/1#issuecomment-505156935


dunglas:
Thanks!

PR #120, opened 2019-07-23
Add support for proxy headers

### LINKED ISSUE #114: mercure.error.log : Client IP wrong
Hello,

I am using Mercury behind a nginx proxy with the following configuration:
    location / {
        proxy_pass http://127.0.0.1:xxxx;
        proxy_read_timeout 24h;
        proxy_http_version 1.1;
        proxy_set_header Connection "";
    }

Unfortunaly, in the logs file "mercure.error.log", the specified IP address is incorrect: client IP is always 127.0.0.1 :(

I tried to add parameters "proxy_set_header", but no success :
    proxy_set_header Host $ host;
    proxy_set_header X-Real-IP $ remote_addr;
    proxy_set_header X-Forwarded-For $ proxy_add_x_forwarded_for;

Can you fix the problem to have the real IP addresses?
Merci par avance Kévin :)

dunglas:
This PR introduces a new `USE_FORWARDED_HEADERS`. When set to `1`, the hub will read the `X-Forwarded-*` or the `Forwarded` headers provided by intermediate proxy servers.

Closes #114 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,7 @@ To install Mercure in a [Kubernetes](https://kubernetes.io) cluster, use the off
 * `READ_TIMEOUT`: maximum duration for reading the entire request, including the body, set to `0s` to disable (default), example: `2m`
 * `SUBSCRIBER_JWT_KEY`: must contain the secret key to valid subscribers' JWT, can be omited if `JWT_KEY` is set
 * `WRITE_TIMEOUT`: maximum duration before timing out writes of the response, set to `0s` to disable (default), example: `2m`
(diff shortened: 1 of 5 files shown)


stof:
Many proxies are using either the `X-Forwarded-*` headers or the `Forwarded` header, but not both in parallel. So the config should allow trusting only one of them.
Same is true for `X-Real-IP` btw.

stof:
Btw, this is exactly the reason why `symfony/http-foundation` requires to set which headers get trusted.

dunglas:
It’s how the Gorilla handler works. You don’t have to set both in parallel, only one between X-Forwarded-For, Forwarded and X-Real-IP is enough.

However, headers must be set set for both the IP, the host and the protocol.

stof:
the thing is, if you set one of them in your proxy and you don't clear the other one (because your proxy just does not support it, which might be out of your control), an attacker could use the other header to inject a fake value (which unsupported proxy header allows such attack depends on the header priority in the Gorilla handler)

dunglas:
Indeed. It’s why we’re documenting that both headers have to be set. Having a more flexible configuration could be nice, but it must be added to Gorilla first. Currently Gorilla doesn’t support this.

PR #121, opened 2019-07-23
Update nginx config with proxy headers

Pyrrah:
Link w/ #120: Add comments for users using argument USE_FORWARDED_HEADERS=1

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,10 @@ server {
         proxy_read_timeout 24h;
         proxy_http_version 1.1;
         proxy_set_header Connection "";
+        ## Remove comments only if you use argument USE_FORWARDED_HEADERS=1 ##
+        #proxy_set_header Host $host;
+        #proxy_set_header X-Real-IP $remote_addr;
+        #proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -322,10 +322,9 @@ server {
         proxy_read_timeout 24h;
         proxy_http_version 1.1;
         proxy_set_header Connection "";
-        ## Remove comments only if you use argument USE_FORWARDED_HEADERS=1 ##
-        #proxy_set_header Host $host;
-        #proxy_set_header X-Real-IP $remote_addr;
-        #proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+        ## Be sure to set USE_FORWARDED_HEADERS=1 to allow the hub to use those headers ##
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -322,6 +322,10 @@ server {
         proxy_read_timeout 24h;
         proxy_http_version 1.1;
         proxy_set_header Connection "";
+        ## Remove comments only if you use argument USE_FORWARDED_HEADERS=1 ##
```
dunglas: It’s safe to always provide these headers, no need to comment them. So, the comment should be “Be sure to set USE_FORWRDED_HEADERS to 1 to allow the hub to use those headers”.

README.md
```diff
@@ -322,6 +322,10 @@ server {
...
+        #proxy_set_header Host $host;
```
dunglas: X-Forwarded-Host must be set to prevent a security vulnerability. There are no needs to override Host however.

README.md
```diff
@@ -322,6 +322,10 @@ server {
...
+        #proxy_set_header X-Real-IP $remote_addr;
```
dunglas: Setting this one is issueless is X-Forwarded-For is set.
However both X-Forwarded-Scheme must be set to prevent an attacker to hijack it.

README.md
Pyrrah: Ok to change the comment with "Be sure to set USE_FORWARDED_HEADERS=1 to allow the hub to use those headers"

README.md
Pyrrah: I withdraw that

README.md
Pyrrah: I withdraw that. Thanks for the feedback

dunglas:
It's not possible to remove the other values. If the `proxy_headers` handler is registered, both `X-Forwarded-For`, `X-Forwarded-Host` and `X-Forwarded-Proto` must be set set (or removed) by the reverse proxy. Otherwise, it introduces a security issue.

Pyrrah:
Thanks, I added the missing headers.
Also, I corrected a mistake in the configuration with port 443 (not 80).

README.md
```diff
@@ -322,6 +322,11 @@ server {
         proxy_read_timeout 24h;
         proxy_http_version 1.1;
         proxy_set_header Connection "";
+
+        ## Be sure to set USE_FORWARDED_HEADERS=1 to allow the hub to use those headers ##
+        proxy_set_header X-Forwarded-Host $host:$server_port;
+        proxy_set_header X-Forwarded-Server $host;
```
dunglas: This one is not supported, but `X-Forwarded-Proto` is missing.

dunglas:
```
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Host $host;
        proxy_set_header X-Forwarded-Proto $scheme;
```

Pyrrah:
ok, I'm tired -_-
I didn't take the right parameter you gave me. It's good now.

dunglas:
Thanks @Pyrrah!

PR #129, opened 2019-07-29
Update README.md with new Learn resources

vitorluis:
Added a reference link to an article with a Laravel example on how to use Mercure


dunglas:
Awesome! Thanks @vitorluis

PR #134, opened 2019-08-13
ReactJs publisher example

DamienVauchel:
Hi,

there is an example of use with ReactJs, Redux and Axios libraries.


examples/publisher-reactjs.js
```diff
@@ -0,0 +1,63 @@
...
+import Axios from 'axios'
+import { connect } from 'react-redux'
+import querystring from 'querystring'
+import store from '../../store'
+
+class MercureTest extends React.Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
```
dunglas: Shouldn't these values be stored in a `const` defined before the class instead of in a state? Also, couldn't we use a functional component instead of a class?

examples/publisher-reactjs.js
```diff
@@ -0,0 +1,63 @@
...
+          return (
+	    <div key={`notif-${res.id}`}>{res.message}</div>
+          )
+        })}
+      </div>
+    )
+  }
+}
+
+const mapStateToProps = state => {
```
toofff:
```
const mapStateToProps = state => ({
  notifications: state.mercure.notifications
})
```

examples/publisher-reactjs.js
```diff
@@ -0,0 +1,63 @@
...
+    })
+  }
+  
+  render() {
+    const { notifications } = this.props
+
+    return (
+      <div>
+        <button type="button" onClick={() => this.handleClick()}>Notify!</button>
+        {notifications.map(res => {
```
toofff:
```
{notifications  && notifications.map(notification => (
  <div key={`notif-${notification.id}`}>{notification.message}</div>
))}
```

examples/publisher-reactjs.js
DamienVauchel: Sure, sorry, I'm new in React and this component is an extract from a component I use in my project (and more complicated than here).

Here, I need componentDidMount function so I don't really see how to make a function insteed of a component with it.

examples/publisher-reactjs.js
```diff
@@ -0,0 +1,57 @@
...
+      })
+    }
+
+    return (
+      <div style={{ padding: '40vh 40vw', color: 'black' }}>
+        <button type="button" onClick={handleClick}>
+          Notify!
+        </button>
+        {notifications && notifications.map(notification => {
+          return <div key={`key-${notification.id}`}>{notification.message}</div>
```
toofff: With add `return`?

this should work
```
{notifications && notifications.map(notification => (
  <div key={`notif-${notification.id}`}>{notification.message}</div>
))}
```

dunglas:
Closing as we'll soon provide a library for React. Thanks for contributing @DamienVauchel!

PR #147, opened 2019-08-28
key-set clarifications

Spomky:
This PR tries to clarify the use of the `key-set` parameter.

* It says that only public keys should be shared (if private or shared keys are shared then it is completely useless).
* It adds an example of JWKSet object
* `encode` => `encrypt`
* `encyption` => `encryption`

Happy to discuss if you have any comments.
BR

### DIFF
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -86,8 +86,8 @@ The publisher **MAY** provide the following target attributes in the Link Header
     resource. Setting the `content-type` attribute is especially useful to hint that partial updates
     will be pushed, using formats such as JSON Patch [@RFC6902] or JSON Merge Patch [@RFC7386].
 
- *  `key-set=<JWKS>`: the key(s) to decrypt updates encoded in the JWKS (JSON Web Key Set) format
-    (see the Encryption section).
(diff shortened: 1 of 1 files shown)


spec/mercure.md
```diff
@@ -86,8 +86,8 @@ The publisher **MAY** provide the following target attributes in the Link Header
     resource. Setting the `content-type` attribute is especially useful to hint that partial updates
     will be pushed, using formats such as JSON Patch [@RFC6902] or JSON Merge Patch [@RFC7386].
 
- *  `key-set=<JWKS>`: the key(s) to decrypt updates encoded in the JWKS (JSON Web Key Set) format
-    (see the Encryption section).
+ *  `key-set=<JWKS>`: the public key(s) encoded in the JWKS (JSON Web Key Set [@!RFC7517]) format.
+     To be used by the publisher to encrypt an update before sending it to the hub (see the Encryption section).
```
dunglas: The idea here is to pass a private key (can be a key used only for the current session for instance) from the publisher to the subscriber, so the subscriber can decrypt the messages, but the Hub (that can be operated by a third-party service) cannot (because the private key is transferred directly from the publisher to the subscriber, without passing by the hub).

Spomky:
Closing as the behaviour is still not clear for me…
We can discuss about this topic [in private if needed](https://gitter.im/Spomky).

PR #149, opened 2019-08-30
Update the demo JWT key to match the default JWT token

emodric:
The default JWT token uses `!ChangeMe!` key, while all commands to start the Mercure hub use `MyJWTKey`, making it impossible to subscribe to updates when following the examples in the readme, instead receiving 401 errors.


dunglas:
Thanks @emodric!

Issue #82, opened 2019-04-29
JWT RS256 support for better integration with ApiPlatform and security


dvc:
Hello everybody!

Shared jwt-secret or static tokens is not so protected as public key on Mercure side. IMHO :)

see also: https://github.com/api-platform/api-platform/issues/1123

dunglas:
Indeed, it would be nice to have support for other algorithms. PR welcome!

Spomky:
>  is not so protected as public key

There is no reason to say that unless you use a hashing function like MD5 or SHA-1.
The advantages with symmetric key algorithms is that they are fast.

Anyway I concur with @dunglas and other algorithms should be supported

dunglas:
See #156

PR #156, opened 2019-09-14
Add Support for RSA Algorithms

SerethiX:
As Requested in #82 and because I want to keep the private key private, I set down and modified the authorization module.


hub/authorization.go
```diff
@@ -71,10 +74,28 @@ func authorize(r *http.Request, jwtKey []byte, publishAllowedOrigins []string) (
...
-		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
-			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
+		switch token.Method.(type) {
+			case *jwt.SigningMethodHMAC:
+				return key, nil
+			case *jwt.SigningMethodRSA:
+				block, _ := pem.Decode(key)
+
+				if block == nil {
+					return nil, errors.New("public key error")
```
dunglas: Shouldn't we return the error returned by `Decode` instead? It will be easier to debug.

hub/authorization.go
```diff
@@ -71,10 +74,28 @@ func authorize(r *http.Request, jwtKey []byte, publishAllowedOrigins []string) (
 // validateJWT validates that the provided JWT token is a valid Mercure token
 func validateJWT(encodedToken string, key []byte) (*claims, error) {
 	token, err := jwt.ParseWithClaims(encodedToken, &claims{}, func(token *jwt.Token) (interface{}, error) {
-		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
-			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
+		switch token.Method.(type) {
```
dunglas: Instead of relying on the algorithm selected by the client, shouldn't we create a new environment variable to configure the algorithm (having `HMAC ` as default) and always use the configured one?

hub/authorization.go
SerethiX: That's the thing: it does not return one:

```go
func Decode(data []byte) (p *Block, rest []byte) {
```

dunglas: Fine then!

hub/authorization.go
SerethiX: The algorithm is not selected by the client, but the identity provider signing the jwt.
IMO we should be able to trust the selected method. It must be compatible with the private key anyways.

dunglas: The client can forge a JWT token and then select the algorithm to use.

https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/

>  In any case, I don't think JWT libraries should even look at the alg field in the header, except maybe to check that it matches what was the expected algorithm.
> Anyone using a JWT implementation should make sure that tokens with a different signature type are guaranteed to be rejected.

There is no reason to trigger a RSA check if the a HMAC key is configured (and vice versa).

SerethiX: You are right, haven't thought of this.

Algo none is forbidden, but forging the public key into an HMAC signed JWT was still present.

Added new commits and rebased my branch.

SerethiX:
> 
> 
> Sorry for the late review... And thank you for this great contribution! I just made a small comment but globally it looks good to me.

Requested changes implemented.

dunglas:
Thank you very much @SerethiX

Viktor-Bredihin:
@SerethiX @dunglas 
any idea how to make it work with symfony bundle https://github.com/symfony/mercure-bundle/issues/39 ?

dunglas:
There is nothing specific related to JWT in the Symfony. Using the proper algorithm when generating the token should be enough.

Issue #160, opened 2019-09-30
Split docs in multiple file


dunglas:
The readme starts to be too big. Splitting it in multiple files would help the readability.

PR #161, opened 2019-10-01
Docs: splitted FAQ.md and HUB_IMPLEMENTATION.md to help readability

aureleoules:
Implements #160


FAQ.md
```diff
@@ -0,0 +1,94 @@
+## FAQ
```
dunglas: Could you use lowercase for the file names please? Thanks for contributing!

aureleoules: It's fixed!

dunglas:
Thanks @aureleoules

PR #171, opened 2019-10-07
Fix path to license

jderusse:


dunglas:
Good catch, thanks Jeremy!

PR #177, opened 2019-10-09
Fix update not dispatched when alternate IRI matches

dunglas:
Thanks to @DamienVauchel for finding this bug!


DamienVauchel:
And I thank you for your fast effective answer. 
I tested this new version and it fixed my problem

PR #179, opened 2019-10-10
Fix possible lost of updates

### LINKED ISSUE #178: Possible lost of update

dunglas:
Close #178


dunglas:
The order isn't guaranteed, by the spec (it's difficult to guarantee order in clustered mode). However we should prevent to send updates twice, so this patch isn't enough.
Maybe could we add a `ready` flag to the subscriber struct, set it to true immediately after `sendMissedUpdate`, and start broadcasting new updates only when `ready` is true. Will probably have to implement some synchronization mechanism too.

dunglas:
Closing in favor of #181, which is a better and cleaner approach!

PR #181, opened 2019-10-13
High Availability in mercure HUB

jderusse:
I'm exited to open that PR which brings **High Availablity** to mercure HUB!

[asciicast](https://asciinema.org/a/u8ZHHQgof7I2YE9nN7iAXkcwd)

## how

By combining HUB with Redis's stream to broadcast Updates to several HUB, it provides several features:
- scale HUB: to handle extra load
- fault resiliency: Shuting down 1 hub is not an issue, others hub in the cluster will take the load

## disclamer 

First of all, I want to **apologie for this HUGE PR**. I tried to perform the smaller change as possible, but, when digging I finally have to break a lot of thinks.
tell me if you prefer several smaller PRs.

### decouple Hub from bolt

NewHub does not return an instance of bolt anymore. Hub is responsible of closing resources it creates.

### deprecate history env variables

Some options depends on the choosen Backend (ie. cleanup_frequency for BOLT). This PR adds a new HISTORY_DSN the parameters HISTORY_X are handled by DSN's querystring.   

### merge History and Dispatcher together

`History` and `Dispatcher` have been deleted, and a new `Stream` interface is created. This interface is in charge of providing cursors who'll stream data from a given point in time (could be now, or a previous EventId)
The HUB is no more responsible of broadcasting updates to subscribers (chans addSubscriber, removeSubscriber, updateSubscriber) this has been delegated to the `stream` which have all information to know if an "update" has been dispatched from history or live streaming and ease the fix of #178

## TODO
* [x] increase code coverage
* [x] reduce amount of goroutine per subscriber
* [ ] allow several URL in redis
* [x] remove sleep in tests
* [x] reduce complexity


hub/subscribe.go
```diff
@@ -17,41 +17,60 @@ func (h *Hub) SubscribeHandler(w http.ResponseWriter, r *http.Request) {
...
-			continue
 		}
-
-		select {
-		case serializedUpdate, open := <-updateChan:
-			if !open {
-				return
+	} else {
+		updates := make(chan *Update)
+		go func() {
```
dunglas: My construct was designed to not spawn a useless goroutine (goroutines can become costly on high traffic services), could you switch back to the previous construct?
Also, it makes the code simpler.

hub/subscribe.go
```diff
@@ -17,41 +17,60 @@ func (h *Hub) SubscribeHandler(w http.ResponseWriter, r *http.Request) {
...
-			serializedUpdate, open := <-updateChan
-			if !open {
-				return
+			update, err := pipe.Read()
+			if err != nil {
+				if err == ErrClosedPipe {
+					return
+				}
+				log.Error(err)
+			} else {
```
dunglas: `continue` could be used to prevent the `else`

hub/subscribe.go
```diff
@@ -17,41 +17,60 @@ func (h *Hub) SubscribeHandler(w http.ResponseWriter, r *http.Request) {
...
+	if h.options.HeartbeatInterval == time.Duration(0) {
+		for {
 			// No heartbeat defined, just block
-			serializedUpdate, open := <-updateChan
-			if !open {
-				return
+			update, err := pipe.Read()
+			if err != nil {
+				if err == ErrClosedPipe {
+					return
```
dunglas: `close(updates)`? Maybe that this code could be factorized with the similar one below?

hub/subscribe_test.go
```diff
@@ -113,39 +113,40 @@ func TestSubscribeNoTopic(t *testing.T) {
...
 				Topics: []string{"http://example.com/hub?topic=faulty{iri"},
 				Event:  Event{Data: "Faulty IRI", ID: "d"},
 			})
-			hub.updates <- newSerializedUpdate(&Update{
+			hub.stream.Write(&Update{
 				Topics: []string{"string"},
 				Event:  Event{Data: "string", ID: "e"},
 			})
 
+			time.Sleep(8 * time.Millisecond)
```
dunglas: it looks suspicious.

hub/subscribe_test.go
```diff
@@ -206,34 +208,35 @@ func TestUnsubscribe(t *testing.T) {
...
 				Event:   Event{Data: "Hello World", ID: "b", Type: "test"},
 			})
-			hub.updates <- newSerializedUpdate(&Update{
+			hub.stream.Write(&Update{
 				Targets: map[string]struct{}{"hello": {}, "bar": {}},
 				Topics:  []string{"http://example.com/reviews/23"},
 				Event:   Event{Data: "Great", ID: "c", Retry: 1},
 			})
 
+			time.Sleep(8 * time.Millisecond)
```
dunglas: Same here and everywhere else

hub/publish_test.go
```diff
@@ -145,21 +146,22 @@ func TestPublishNotAuthorizedTarget(t *testing.T) {
 func TestPublishOK(t *testing.T) {
 	hub := createDummy()
 
+	pipe, err := hub.stream.CreatePipe("")
+	assert.Nil(t, err)
```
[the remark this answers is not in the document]
jderusse: assert report an error and continue the execution.
require report an error and stop the test.

I use "require" to avoid null pointer panic when the following assertions expect a well known object with methods and properties.

Here, I require an `pipe` because following code is calling `pipe.write()`, but just want to assert that err is not nil. It's fine if the test continue to report other errors.

hub/stream.go
```diff
@@ -0,0 +1,116 @@
...
+// Write pushes updates in the Stream
+func (s *LiveStream) Write(update *Update) error {
+	select {
+	case <-s.done:
+		return ErrClosedStream
+	default:
+	}
+
+	var closedPipes []*Pipe
+	err := func() error {
```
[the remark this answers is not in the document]
jderusse: yes, to not forget to release the lock if the code is updated to add another `return`.

If you don't like, I can remove it

hub/publish_test.go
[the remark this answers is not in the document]
dunglas: Ok makes sense.

hub/stream.go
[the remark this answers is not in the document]
dunglas: Yes please, let's keep the code as simple as possible, this trick is fine, but I prefer to use it only when it allows to remove extra conditions.

hub/subscribe.go
jderusse: Because of the `select` (to create a race between updates and heartbeat), I have to create an intermediate chan.

For the record, the object Pipe is just a Chan that gracefully handle Closing (reader and writer can close the chan) while a standard chan will panic if someone try to write on it.

I've 3 options:
- keep this goroutine
- add a method `ReadChan` in Pipe that return the chan
- remove the Pipe thing, `Stream::CreatePipe` returns a chan, and add a method `Stream::CloseChan` to gracefully close the chan from reader and writers

hub/subscribe.go
jderusse: the chan `updates` does not exists here. There is no race with heartbeat, we can directly read from block `Read` method.

.env
```diff
@@ -5,7 +5,7 @@ ALLOW_ANONYMOUS=1
 #CERT_FILE=fixtures/tls/server.crt
 #KEY_FILE=fixtures/tls/server.key
 CORS_ALLOWED_ORIGINS=http://localhost:3000,http://localhost:8000
-DB_PATH=
+HISTORY_DNS=bolt://update.db
```
dunglas:
```suggestion
HISTORY_DSN=bolt://update.db
```

What do you think about `STREAM_URL` (similar to Heroku/Symfony `DATABASE_URL`, or even `ADAPTER_URL` (adapter is the term used by socketio), because it's not only about history.  Or `TRANSPORT_URL`?

If we choose `adapter` instead of `stream`, we should also rename the interface.

hub/server.go
```diff
@@ -58,7 +58,7 @@ func (h *Hub) Serve() {
 			h.server.TLSConfig = certManager.TLSConfig()
 
 			// Mandatory for Let's Encrypt http-01 challenge
-			go http.ListenAndServe(":http", certManager.HTTPHandler(nil))
+			go http.ListenAndServe(h.options.AcmeAddr, certManager.HTTPHandler(nil))
```
dunglas: I'm not sure it is worth it, by the spec the acme HTTP challenge can only be resolved on the `80` port. At least, we should find a better option name, such as `AcmeHTTP01Addr`.

.env
jderusse: You're right, I first kept `HISTORY_` for consistency with other parameter. But once we removed them, it's not necessary anymore.
I don't like `adapter` because it's too generic and could be anything.

The parameter should be understandable from a user POV, `STREAM_URL` may not talk to everybody. I'll note

IMHO The parameter don't have the same prefix as the interface (like `MAILER_DNS` for `Transport` interface in symfony)

hub/server.go
jderusse: This change had 2 purposes: 
- being able to test that part (we can't start a listener on port 80 without running test-suite with root user)
- allows several hubs running on the same server. Off course, a loadbalancer/proxy on top of the server should  route the traffic to the right port

jderusse: you may prefer a dedicated PR?

jderusse: moved in #192

telemmaite:
As far as I see, HA can be achieved by Clustering, which is the PRO+ option in the managed Mercure version.

What are the real benefits and needs to connect multiple hubs via Redis instead of boltDB.?
On the other end, how could HA be achieved with Clustering when the boltDB/localfiles are used?

jderusse:
You cannot have HA with a single node in your infra.

You cannot setup a HUB cluster with bolt, because bolt use a local file storage, you have to use a "shared" database: Redis, Kafka, Kinesis, PgSLQ.

If you setup 2 Hub with bolt, you'll have 2 distinguished/isolated instances: Consumer connected to 1 instance won't receive messages published by producers connected to another instances.

I'll probably rebase this PR soon, using plugin does not work well with golang, and hosting a clone/fork of this project in my namespace does not make sens.

telemmaite:
> You cannot setup a HUB cluster with bolt, because bolt use a local file storage, you have to use a "shared" database: Redis, Kafka, Kinesis, PgSLQ.
> 
> If you setup 2 Hub with bolt, you'll have 2 distinguished/isolated instances: Consumer connected to 1 instance won't receive messages published by producers connected to another instance.

So we can agree, that the managed version of Mercure is not only about hosting & running multiple instances of the hub (clustering) but it also includes HA setup by using, probably, any of the above databases? 

---
We are big users of **Symfony** framework on multiple fronts, and we would like to go with the **Symfony/Mercure bundle** but real stopping stone is HA and how to achieve it with Mercure in **K8s** environment, ie. multiple pods shared **Redis**. 
Currently, this Pull requests the only option I could consider as viable or we/my company have to move back to **Socket.io with Redis adapter for HA**, which is solid and tested on our end.

telemmaite:
Hey guys any update on this? I feel like on a crossroad.

progmancod:
Any news?

martijnhartlief:
Any updates on this?

dunglas:
The HA version now includes a Redis Transport (as well as transports using PostgreSQL, Kafka and Pulsar): https://mercure.rocks/docs/hub/cluster

I'm closing this PR as the refactoring introduced by Jeremy has been merged, and HA transports will not be included in the official FOSS distribution (however, it's now easy, thanks to this refactoring, to create custom distributions including additional transports).

PR #185, opened 2019-10-18
Mac OS Catalina Installation error -> add to note

castroCrea:


dunglas:
Hi, thanks for contributing but could you take a screenshot in English and add it to the repo?

dunglas:
A simple text description would do the trick too.

dunglas:
Thanks @castroCrea!

Issue #182, opened 2019-10-15
Change the default URL of the hub to /.well-known/mercure


dunglas:
It will prevent conflicts with existing URLs and improve the discoverability mechanism.

dunglas:
Host meta could be interesting too: https://tools.ietf.org/html/rfc6415

PR #189, opened 2019-10-21
Change the default URL of the hub

jrwren:
to /.well-known/mercure

### DIFF
--- a/LoadTest.scala
+++ b/LoadTest.scala
@@ -19,7 +19,7 @@ import scala.util.Properties
 
 class LoadTest extends Simulation {
   /** The hub URL */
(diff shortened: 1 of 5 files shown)


jrwren:
for #182

dunglas:
Hi @jrwren, thanks for this! Code changes are ok, but the spec must also be updated to indicate the URL `SHOULD` be `/.well-known/mercure`

jrwren:
I looked over the spec and I don't know know where to put this information.

dunglas:
I'll take care of this! Thanks

dunglas:
Thanks @jrwren

Issue #187, opened 2019-10-21
Use viper for the config


dunglas:
https://github.com/spf13/viper

It will allow to configure Mercure using env vars but also YAML, TOML, JSON...

PR #190, opened 2019-10-23
Added viper for config resolution

Herocod3r:
PR for this [Issue](https://github.com/dunglas/mercure/issues/187#issue-510302398)


.gitignore
```diff
@@ -9,3 +9,4 @@
 /spec/mercure.xml
 __pycache__
 *.pyc
+.idea/
```
dunglas: Can you remove this? It should be in a global gitignore file.

config/app_config.go
```diff
@@ -0,0 +1,32 @@
...
+	"os"
+	"path"
+)
+
+func init(){
+	rootPath,_ := os.Getwd()
+	viper.AutomaticEnv()
+	viper.AddConfigPath(".")
+	viper.AddConfigPath( path.Join(rootPath,".."))
+	err := viper.ReadInConfig()
```
dunglas: Maybe something like this?

```suggestion
    if err := viper.ReadInConfig(); err != nil {
        if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
            // log.Panic(err)
        }
    }
```

config/app_config.go
```diff
@@ -0,0 +1,32 @@
+package config
```
dunglas: Couldn't this code be moved directly in `hub/options.go`?

.gitignore
Herocod3r: Alright, will do

config/app_config.go
Herocod3r: True, this makes sense

config/app_config.go
Herocod3r: This would imply that config is tied to only the *hub* package? my thoughts behind this was an independent package who handles configuration and could be used anywhere ..say we need to somehow add another package in the future.. don't you think dependency on the *hub* package is somewhat bringing unnecessary baggage ?

dunglas: But as the hub package depends of the config package that depends of viper etc, it doesn't fix this issue right now, isn't it?

config/app_config.go
```diff
@@ -0,0 +1,38 @@
...
+
+func init(){
+	setupConfig()
+}
+
+func setupConfig(){
+	rootPath,_ := os.Getwd()
+	viper.AutomaticEnv()
+	viper.AddConfigPath(".")
+	viper.AddConfigPath( path.Join(rootPath,".."))
```
jderusse: sounds like an arbitrary path.
What's about
- `.`
- `/etc/mercure/`
- `$HOME/mercure`

config/app_config.go
```diff
@@ -0,0 +1,38 @@
...
+	if err := viper.ReadInConfig(); err != nil {
```
jderusse: shouldn't you define a config name? `viper.SetConfigName("mercure") `

config/app_config.go
```diff
@@ -0,0 +1,38 @@
...
+		log.Println("Loading config from enviroment variables..")
+	}
+}
+
+func Get(key string) interface{}{
+	return viper.Get(key)
+}
+
+func GetString(key string) string{
+	value := Get(key)
```
jderusse: `return viper.GetString(key)`

dunglas:
Hi @Herocod3r. I rebased and improved your PR in https://github.com/dunglas/mercure/pull/203. Could you take a look?

PR #191, opened 2019-10-24
Added JWT_KEY_FILE, PUBLISHER_JWT_KEY_FILE and SUBSCRIBER_JWT_KEY_FILE

Richard87:
To simplify using key files

When running Mercure from Docker-compose, it's tricky with multiline env-variables


docs/hub/config.md
```diff
@@ -47,7 +50,7 @@ bCd7nPuNAyYHCOOHAgMBAAE=
 
 Bash
 ```
-JWT_KEY=`cat jwt_key.pub` ./mecure
```
jderusse: This line of code shows how to use file instead of env variable.

Richard87: Thanks, I'm aware, but that doesn't work with docker-compose, or in my second attempt with Intellij (the problem here was that Intellij replaces newline with space, and golang's pem.Decode failes if the key doesn't have newlines in it in the regular format)

jderusse: did you try to change the command (within docker-compose) in `CMD: 'JWT_KEY=$$(cat jwt_key.pub) mercure'` (maybe some things need to be escaped) or to add `export JWT_KEY=$(cat jwt_key.pub)` in the entrypoint,

Richard87: Good idea and I just tried it, but since the finale mercure image is built from the `scratch` image, it doesn't have the cat binary :(

Richard87: Also, I feel Mercure should be able to load a key from a file, and resorting to bash "hacks" should be avoided :)

hub/options.go
```diff
@@ -36,13 +37,36 @@ type Options struct {
 	Demo                    bool
 }
 
-func getJWTKey(role string) string {
-	key := os.Getenv(fmt.Sprintf("%s_JWT_KEY", role))
-	if key == "" {
-		return os.Getenv("JWT_KEY")
+func getJWTKey(role string) (string, error) {
+	roleKeyFile := os.Getenv(fmt.Sprintf("%s_JWT_KEY_FILE", role))
```
dunglas:
```suggestion
	jwtKeyFile := os.Getenv(fmt.Sprintf("%s_JWT_KEY_FILE", role))
```

hub/options.go
```diff
@@ -36,13 +37,36 @@ type Options struct {
...
+		return string(file), err
+	}
+
+	roleKey := os.Getenv(fmt.Sprintf("%s_JWT_KEY", role))
+	if roleKey != "" {
+		return roleKey, nil
+	}
+
+	keyFile := os.Getenv("JWT_KEY_FILE")
+	if keyFile != "" {
```
dunglas: Extract this code in a function, so it doesn't have to be duplicated

hub/options.go
```diff
@@ -36,13 +37,36 @@ type Options struct {
...
+		file, err := ioutil.ReadFile(keyFile)
+		if err != nil {
+			return "", fmt.Errorf("error loading file: %v", err)
+		}
+		return string(file), err
 	}
 
-	return key
+	key := os.Getenv("JWT_KEY")
```
dunglas: Useless, can just be return `return os.Getenv("JWT_KEY"), nil`

Richard87:
I'm on it :)

Richard87:
All requested changes have been fixed, and tests have been added :)

Richard87:
I'm closing it because it's to different than the new Viper-based config so it can't be easily refactored.... I will open a new pull-request based on v.0.8.0...

flaugere:
@Richard87 @dunglas 
Hi,
I'm trying to use docker secrets (Swarm mode) and I could not find any more clues to implement in that way. This PR seems to but I don't really understand the reason why it is closed.

Richard87:
Hi! I closed it because of a major change to the master branch made my code incompatible, and way harder to implement. And because of unrelated changes, i didn't need this, so I havent worked on it again :)

flaugere:
@Richard87 
Thanks for the explanation 👍

PR #192, opened 2019-10-27
Add Acme HTTP01 addr configuration

jderusse:
This PR allows to configure Acme HTTP01 resolver listening in a non-standard port.
It had several benefits:
- adds test on this part: Be listening on a port >1024, it allows a non-root user starting the service
- allows several HUB on the same machine: A proxy/loadbalancer should be setup to forward the traffic to the right port 
- starts the hub with a no-root user: similar to first benefit in testing mode, this allows to start the service with a non-root user and use a proxy (like in second benefit) to root traffic to the right port


hub/server_test.go
```diff
@@ -164,3 +166,47 @@ func TestServe(t *testing.T) {
...
+		Compress:               true,
+	})
+
+	h.Start()
+	go func() {
+		h.Serve()
+	}()
+
+	// let time to the acme server to start
+	time.Sleep(100 * time.Millisecond)
```
dunglas: Could you replace this by a wait?

jderusse: replaced by a loop bellow

dunglas:
Thanks!

PR #193, opened 2019-10-27
Refactor interfaces to ease Plugin integration

jderusse:
This PR is a split of #181. It removes the "redis" stream and focus on refactoring the HUB to support such architecture.

It will ease to create a plugin system which allows initiatives like `redis_stream` or `kafka_stream` in dedicated repositories.

TODO: 
* [ ] find a suitable name to the DSN env variable ... (currently `TRANSPORT_URL`, could be `STREAM_URL`, `BACKEND_URL`, ...)


hub/pipe.go
```diff
@@ -0,0 +1,55 @@
...
+	done    chan struct{}
+}
+
+// NewPipe creates pipes
+func NewPipe() *Pipe {
+	return &Pipe{make(chan *Update, 1), make(chan struct{})}
+}
+
+// Write pushes updates in the pipe. Returns true is the update is pushed, false otherwise.
+func (c *Pipe) Write(update *Update) bool {
```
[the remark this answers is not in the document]
jderusse: Because this class was previously named `cursor` :-p

hub/stream.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if options.TransportURL.Scheme == "bolt" {
+		return NewBoltStream(options)
+	}
+
+	return nil, fmt.Errorf(`no stream available for "%s"`, options.TransportURL)
+}
+
+// LiveStream implements the StreamInterface without database and simply broadcast the live Updates
+type LiveStream struct {
```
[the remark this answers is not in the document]
jderusse: renamed

hub/subscribe_test.go
```diff
@@ -113,39 +113,40 @@ func TestSubscribeNoTopic(t *testing.T) {
...
 				Topics: []string{"http://example.com/hub?topic=faulty{iri"},
 				Event:  Event{Data: "Faulty IRI", ID: "d"},
 			})
-			hub.updates <- newSerializedUpdate(&Update{
+			hub.stream.Write(&Update{
 				Topics: []string{"string"},
 				Event:  Event{Data: "string", ID: "e"},
 			})
 
+			time.Sleep(8 * time.Millisecond)
```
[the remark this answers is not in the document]
jderusse: Tried hard, but didn't figured out how... :( help wanted  (same bellow)

hub/bolt_transport_test.go
```diff
@@ -0,0 +1,247 @@
...
+	require.NotNil(t, transport)
+	transport.Close()
+
+	url, _ = url.Parse("bolt://")
+	_, err = NewBoltTransport(&Options{TransportURL: url})
+	assert.EqualError(t, err, `invalid bolt "bolt:" dsn: missing path`)
+
+	url, _ = url.Parse("bolt:///root/test.db")
+	_, err = NewBoltTransport(&Options{TransportURL: url})
+	assert.EqualError(t, err, `invalid bolt "bolt:///root/test.db" dsn: open /root/test.db: permission denied`)
```
dunglas: This test fail on Mac OS:

        	Error:      	Error message not equal:
        	            	expected: "invalid bolt \"bolt:///root/test.db\" dsn: open /root/test.db: permission denied"
        	            	actual  : "invalid bolt \"bolt:///root/test.db\" dsn: open /root/test.db: no such file or directory"

dunglas: Using `/` instead of `/root` fixes the issue.

hub/subscribe_test.go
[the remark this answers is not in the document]
dunglas: It's why the `Start()` method was useful IIRC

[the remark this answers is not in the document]
dunglas: Maybe could we provide a test transport closing the connection after a specified number of messages?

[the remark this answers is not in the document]
dunglas: Shouldn't `transport.Close()` ensure that everything is sent before closing the pipes?

hub/subscribe.go
```diff
@@ -17,41 +18,57 @@ func (h *Hub) SubscribeHandler(w http.ResponseWriter, r *http.Request) {
...
-			f.Flush()
+			if err == context.DeadlineExceeded {
+				// Send a SSE comment as a heartbeat, to prevent issues with some proxies and old browsers
+				fmt.Fprint(w, ":\n")
+				f.Flush()
+
+				continue
+			}
+
+			log.Error(err)
```
dunglas: This line isn't tested, it could be tested using a cancellable context in the test.

hub/subscribe.go
```diff
@@ -87,25 +104,21 @@ func (h *Hub) initSubscription(w http.ResponseWriter, r *http.Request) (*Subscri
 	authorizedAlltargets, authorizedTargets := authorizedTargets(claims, false)
 	subscriber := NewSubscriber(authorizedAlltargets, authorizedTargets, topics, rawTopics, templateTopics, retrieveLastEventID(r))
 
-	if subscriber.LastEventID != "" {
-		h.sendMissedEvents(w, r, subscriber)
+	pipe, err := h.transport.CreatePipe(subscriber.LastEventID)
+	if err != nil {
```
dunglas: This condition is not tested, it can be tested by using a test transport that returns an error

hub/bolt_transport.go
```diff
@@ -0,0 +1,239 @@
...
+		err         error
+		closedPipes []*Pipe
+	)
+
+	t.RLock()
+
+	err = t.persist(update)
+
+	for pipe := range t.pipes {
+		if !pipe.Write(update) {
```
jderusse: should we write the update even if the persist failed?

hub/bolt_transport.go
```diff
@@ -0,0 +1,239 @@
...
+	return pipe, nil
+}
+
+// Close closes the Transport
+func (t *BoltTransport) Close() error {
+	select {
+	case <-t.done:
+		// Already closed. Don't close again.
+	default:
+		t.RLock()
```
jderusse: could be `t.Lock`

hub/bolt_transport.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+		close(t.done)
+		t.db.Close()
+	}
+
+	return nil
+}
+
+// cleanup removes entries in the history above the size limit, triggered probabilistically
+func (t *BoltTransport) cleanup(bucket *bolt.Bucket, lastID uint64) error {
```
jderusse: I wonder if that method shouldn't run in a goroutine called every x seconds

hub/publish.go
```diff
@@ -9,22 +9,14 @@ import (
...
-
-// Publish publish an update locally
-func (*localPublisher) Publish(h *Hub, u *Update) error {
+func (h *Hub) dispatch(u *Update) error {
 	if u.ID == "" {
 		u.ID = uuid.Must(uuid.NewV4()).String()
 	}
 
-	h.DispatchUpdate(u)
+	if err := h.transport.Write(u); err != nil {
```
jderusse: `return h.transport.Write(u)`

hub/bolt_transport.go
dunglas: Good question. Both behaviors have benefits and drawbacks...

hub/bolt_transport.go
dunglas: why not, but maybe in another PR?

hub/bolt_transport.go
jderusse: I see 2 drawback:
- People consuming live streaming won't get the same data as people subscribing later.
- The Hub will return an error to the producer while update has been pushed to the subscribers....

dunglas: Indeed if it we send an error to the publisher, we must not dispatch the update. Let's change.

jderusse:
LGTM

dunglas:
Thank you very much for this @jderusse!

PR #200, opened 2019-11-02
Fix possible lost of update

### LINKED ISSUE #178: Possible lost of update

dunglas:
Close #178. Partial alternative to #181.

### DIFF
--- a/hub/subscribe.go
+++ b/hub/subscribe.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


hub/subscribe.go
```diff
@@ -87,16 +87,21 @@ func (h *Hub) initSubscription(w http.ResponseWriter, r *http.Request) (*Subscri
...
-
 	// Create a new channel, over which the hub can send can send updates to this subscriber.
 	updateChan := make(chan *serializedUpdate)
 
 	// Add this client to the map of those that should
 	// receive updates
+
+	useHistory := subscriber.LastEventID != ""
+	if useHistory {
+		h.history.Lock()
```
jderusse: this locks all writes until this subscriber reads the last missed event. This can be a performance issue if the history is huge or network with that subscriber is slow.

hub/subscribe.go
```diff
@@ -87,16 +87,21 @@ func (h *Hub) initSubscription(w http.ResponseWriter, r *http.Request) (*Subscri
...
+		h.sendMissedEvents(w, r, subscriber)
+	}
 	h.newSubscribers <- updateChan
+	if useHistory {
+		h.history.Unlock()
```
jderusse: are you sure that at this point the new subscriber is fully registered in the hub and pending writes will be dispatched to it?

hub/subscribe.go
jderusse: and that also prevents 2 concurrent reads

PR #201, opened 2019-11-02
Update the spec and the docs to use /.well-known/mercure

### LINKED ISSUE #182: Change the default URL of the hub to /.well-known/mercure

dunglas:
Close #182, follows #189.


Yggdrasil:
Hi @dunglas, this name hasn't been registered with the official "Well-Known URI" registry yet right? https://tools.ietf.org/html/rfc8615#section-5.1

dunglas:
No indeed, I have to do it

PR #202, opened 2019-11-02
Another try to fix #178

### LINKED ISSUE #178: Possible lost of update

dunglas:
Another try to fix #178. Maybe a small risk of duplicate updates.


hub/subscribe.go
```diff
@@ -87,16 +87,19 @@ func (h *Hub) initSubscription(w http.ResponseWriter, r *http.Request) (*Subscri
...
-
 	// Create a new channel, over which the hub can send can send updates to this subscriber.
 	updateChan := make(chan *serializedUpdate)
 
+	h.RLock()
 	// Add this client to the map of those that should
 	// receive updates
+	lastSentEventID := h.lastSentEventID
 	h.newSubscribers <- updateChan
+	h.RUnlock()
```
jderusse: there is still one concurency issue.
if a write is locked by this lock, the event will not be published to the subscriber because:
the event is sent after `lastSentEventID` and the subscriber is not yet registered

PR #203, opened 2019-11-05
Finish 190

### LINKED ISSUE #187: Use viper for the config

dunglas:
Finishes https://github.com/dunglas/mercure/pull/190. Maybe be can we go one step further and completely drop the `Options` struct in favor of Viper itself. It will allow to benefit from all Viper's features including [live reloading](https://github.com/spf13/viper#watching-and-re-reading-config-files) (change a config file or an env var such as the JWT key, and the change is taken into account instantly, without having to reload Mercure). WDYT?

Closes #187, closes #190.


dunglas:
Thanks @Herocod3r

Issue #175, opened 2019-10-08
[Symfony] Not able to setup HTTPS 


Xutyr:
Hi,
I'm not able to use the mercure HUB with HTTPS whereas everything worked fine with HTTP.

**What I want to do :**
I'm trying to move my local env from HTTP to HTTPS. Regarding Mercure I updated my mercure configuration following the document but I should missed something.

**What I did :**

I update my mercure configuration as below:
```
set JWT_KEY=aVerySecretKey
set ACME_HOSTS=mysite.fr
set ALLOW_ANONYMOUS=1
set CORS_ALLOWED_ORIGINS=https://mysite.fr
.\mercure.exe
```
They I update my symfony .env file as below:
```
# See https://symfony.com/doc/current/mercure.html#configuration
MERCURE_PUBLISH_URL=https://mysite.fr/hub
# The default token is signed with the secret key: !ChangeMe!
MERCURE_JWT_KEY=aVerySecretKey
```
Regarding my JS file I set up the connection as below:
```
var handleMercure = () => {
        const url = new URL('https://mysite.fr/hub');
        url.searchParams.append('topic', "https://mysite.fr/notif");
        const eventSource = new EventSource(url, {withCredentials: true});
        eventSource.onmessage = e => {
            console.log(e);
            /* DO SOMETHING */
        }
    };
```
Following Symfony document about mercure and Grafikart video I setup a MercureCookieGenerator.php:

```
class MercureCookieGenerator
{
    /**
     * @var string
     */
    private $secret;

    public function __construct(string $secret)
    {
        $this->secret = $secret;
    }

    public function generate (User $user){
        $token = (new Builder())
            ->withClaim('mercure', ['subscribe' => ["https://mysite.fr/notif/user/{$user->getId()}"]])
            ->getToken(new Sha256(), new Key($this->secret));
        return "mercureAuthorization={$token}; Path=/hub; HttpOnly;";
    }
}
```
And finally I have my own NotificationService.php used to push to mercure a new message:
```
public function sendNotif(?User $User=null, $notif= "", $url="",$messageName="",$messageParam = [], AlertNotification $Alert=null){
        $target = [];
        if($User!==null){
            $target = ["https://mysite.fr/notif/user/{$User->getId()}"];
        }
        $update = new Update("https://mysite.fr/notif", json_encode(['url' => "${url}",'notif'=> "${notif}", 'messageName' => "${messageName}", 'messageParam' => json_encode($messageParam)]), $target);
[...]

Issue #130, opened 2019-08-01
AWS EB - Certificate signed by unknown authority


brutalnv:
Hello!
I faced this issue running mercure on AWS ElasticBeanstalk instance:

```
-------------------------------------
/var/log/eb-docker/containers/eb-current-app/eb-771cc6604e34-stdouterr.log
-------------------------------------
time="2019-08-01T19:42:27Z" level=info msg="Mercure started" protocol=https
2019/08/01 19:43:44 http: TLS handshake error from 31.132.178.66:6018: Get https://acme-v01.api.letsencrypt.org/directory: x509: certificate signed by unknown authority
2019/08/01 19:43:58 http: TLS handshake error from 31.132.178.66:6017: acme/autocert: missing certificate
```

My Dockerrun file:

```
{
 "AWSEBDockerrunVersion": "1",
 "Image": {
  "Name": "dunglas/mercure"
 },
 "Ports": [
	 {
	  "HostPort": "80",
	  "ContainerPort": "80"
	 },
	 {
	  "HostPort": "443",
	  "ContainerPort": "443"
	 }
 ]
}
```

Any ideas what the cause might be?
I'm currently using a workaround by mapping volume with manually generated certs to the docker container and it works like a charm.

dunglas:
It looks like a proxy or a security rule (in Beanstalk?) prevents the hub to contact Let's Encrypt. Another cause can be that the default TLS cert bundles isn't installed, but if you use the official Docker image it shouldn't be the case. Sorry I don't know Beanstalk very well and I never tried Mercure on it.

byhoratiss:
I'm using Mercure as a service with AWS ECS.
I'm using an AWS SSL certificate (provided by ACM), linked to an LB.
My LB is receiving HTTP & HTTPS connections, but only using HTTP to Mercure service running on HTTP only, and you must use the HEARTBEAT environment variable to avoid problems in Mercure service.

I hope I explained myself correctly, let me know if you need some help.

brutalnv:
Thanks for your advice, will try to deploy using ECS and load balancer. Looks like it will be good choice in my this case.

Richard87:
Hi, I got the same problem, but posted in on this issue #175

Richard87:
The problem is the docker image, it is missing the ca-bundle...
Adding thins in the final image in Dockerfile solves the problem:

`
COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
`

[...]

PR #204, opened 2019-11-06
Added ca-bundle to final image (requried by ACME)

Richard87:
Added ca-bundle to final image, this is required for ACME to work :)

Solves #175  and #130


dunglas:
Thanks @Richard87, good catch!

Richard87:
Thanks, my pleasure 😊

PR #216, opened 2019-11-21
Added link to a new Java library to publish messages

vitorluis:


dunglas:
Awesome, thank you @vitorluis

PR #226, opened 2019-12-27
Create Procfile for Heroku deployment

Spomky:
Hi,

In regards to the answer from @dunglas and my comment on [ths SO question](https://stackoverflow.com/questions/56746167/is-this-possible-to-get-mercure-protocol-with-heroku-using-api-platform), I propose you to integrate that `Procfile` to ease the deployment with Heroku.

I am not sure you agree with this PR as it couples the package with Heroku config.
In case you refuse it, I could propose you to add in a dedicated section of the documentation.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: ADDR=":${PORT}" mercure


alouini333:
I created a project for Dokku. I don't know if it's the same for Heroku
https://github.com/alouini333/Deploy-Mercure-Dokku

Spomky:
Hi @alouini333,

It looks very similar. The only difference is related to the listened port that is mandatory for Heroku.

dunglas:
IMO this should be as part of the documentation instead of in the repo, for two reasons:

* Heroku needs the binary, not the source
* It will ends by a proliferation of files at the root of the repo (see #92 for instance)
* Now that the SaaS version is publicly available (breaking, the official announcement hasn't been made yet, but it's online https://mercure.rocks/pricing) I obviously prefer to promote the service that funds the project

Could you add a new docs entry, such as `hub/deployment.md` with a section for Heroku containing this file? We'll do the same for Scalingo.

Spomky:
Hi @dunglas,

Sure I will modify that PR and add this to the doc instead the dedicated Procfile.

> Now that the SaaS version is publicly available

That’s a good news and will be very useful!

dunglas:
Closing for now.

Issue #117, opened 2019-07-15
A way to know how many clients are connected to the hub ?


ramrami:
Is there a way to know how many clients are connected to the hub ? thks

edit:
Use case example: I want to display the number of connected users on a live stream

dmouse:
This looks similar to #28

Issue #100, opened 2019-06-11
Websub protocol


gedw99:
There is a golang websub implementation I saw. Gotta find it. This is designed for server to server.

Am thinking about also supporting websub on top of Mercure. It would be a different entry point but can use all the event tracking still to provide a architecture for microservices.

What do you think ?

dunglas:
Adding a WebSuch endpoint to the hub is something I have in mind since the very beginning. Such addition would be very welcome.

gedw99:
Great to hear.

There was only one golang implementation in GitHub.

https://github.com/adamsanghera/go-websub

dunglas:
The protocol is very simple, maybe could we just ship our own implementation.

gedw99:
Thats lucky and if you want to have a ho at starting it I can help extend from your base.

divine:
> The protocol is very simple, maybe could we just ship our own implementation.

@dunglas well, first of all thank you for the great work, secondly any ETA for this feature anytime soon? Thanks.

divine:
@dunglas any update on this please?

dunglas:
It will be available before the end of 2020. Sorry I don't have a more precise ETA yet, but I've started working on this (design part for now).

jason-dll:
@dunglas Is there any news ?

Thank you for all your work

dunglas:
It's the next entry in my todo list after releasing 2.6 stable!

colinmcdermott:
Any update on this?

PR #238, opened 2020-02-18
Dispatch connection and disconnection events

### LINKED ISSUE #28: Connection events
Is it possible to trigger events and send them to my application server to know if a user is connected/disconnected?

dunglas:
This PR adds a section in the spec and a new option to dispatch updates every time a subscription is created or terminated. It is useful when subscribers have to to be aware of the other currently connected subscribers.

When #100 will be implemented, this feature will also allow servers to be notified when a user is online or offline, which is a frequent enhancement request.

Closes #28, #117.


hub/subscribe.go
```diff
@@ -202,3 +233,53 @@ func (h *Hub) cleanup(s *Subscriber) {
...
 }
+
+func (h *Hub) dispatchSubscriptionUpdate(topics, encodedTopics []string, connectionID string, claims *claims, active bool, address string) {
+	if !h.config.GetBool("dispatch_subscriptions") {
+		return
+	}
+
+	for k, topic := range topics {
+		connection := &subscription{
+			ID:      "https://mercure.rocks/subscriptions/" + encodedTopics[k] + "/" + connectionID,
```
Darkweak: Put `https://mercure.rocks/subscriptions` or at least `https://mercure.rocks` in constant or in env var setup

hub/subscribe.go
```diff
@@ -202,3 +233,53 @@ func (h *Hub) cleanup(s *Subscriber) {
...
+		}
+
+		json, err := json.MarshalIndent(connection, "", "  ")
+		if err != nil {
+			panic(err)
+		}
+
+		u := &Update{
+			Topics:  []string{connection.ID},
+			Targets: map[string]struct{}{"https://mercure.rocks/targets/subscriptions": {}, "https://mercure.rocks/targets/subscriptions/" + encodedTopics[k]: {}},
```
Darkweak: Same here

hub/subscribe_test.go
```diff
@@ -329,6 +333,90 @@ func TestSubscribeTarget(t *testing.T) {
...
+	hub.config.Set("subscriptions_include_ip", true)
+
+	var wg sync.WaitGroup
+	ctx1, cancel1 := context.WithCancel(context.Background())
+	ctx2, cancel2 := context.WithCancel(context.Background())
+	wg.Add(3)
+	go func() {
+		// Authorized to receive connection events
+		defer wg.Done()
+		req := httptest.NewRequest("GET", defaultHubURL+"?topic=https://mercure.rocks/subscriptions/{topic}/{connectionID}", nil).WithContext(ctx1)
```
Darkweak: Use domain or base url constant here

hub/subscribe.go
```diff
@@ -202,3 +233,53 @@ func (h *Hub) cleanup(s *Subscriber) {
...
+
+func (h *Hub) dispatchSubscriptionUpdate(topics, encodedTopics []string, connectionID string, claims *claims, active bool, address string) {
+	if !h.config.GetBool("dispatch_subscriptions") {
+		return
+	}
+
+	for k, topic := range topics {
+		connection := &subscription{
+			ID:      "https://mercure.rocks/subscriptions/" + encodedTopics[k] + "/" + connectionID,
+			Type:    "https://mercure.rocks/Subscription",
```
soyuka: Do you think that the hostname could be made configurable in the future?

hub/subscribe_test.go
```diff
@@ -329,6 +333,90 @@ func TestSubscribeTarget(t *testing.T) {
...
+		req.AddCookie(&http.Cookie{Name: "mercureAuthorization", Value: createDummyAuthorizedJWT(hub, subscriberRole, []string{"https://mercure.rocks/targets/subscriptions"})})
```
Darkweak: Same here

hub/subscribe_test.go
```diff
@@ -329,6 +333,90 @@ func TestSubscribeTarget(t *testing.T) {
...
+		assert.Contains(t, bodyContent, `data:   "subscribe": []`)
+		assert.Contains(t, bodyContent, `data:   "active": true,`)
+		assert.Contains(t, bodyContent, `data:   "active": false,`)
+		assert.Contains(t, bodyContent, `data:   "address": "`)
+	}()
+
+	go func() {
+		// Not authorized to receive connection events
+		defer wg.Done()
+		req := httptest.NewRequest("GET", defaultHubURL+"?topic=https://mercure.rocks/subscriptions/{topic}/{connectionID}", nil).WithContext(ctx2)
```
Darkweak: Same here

hub/subscribe_test.go
```diff
@@ -329,6 +333,90 @@ func TestSubscribeTarget(t *testing.T) {
...
+		req.AddCookie(&http.Cookie{Name: "mercureAuthorization", Value: createDummyAuthorizedJWT(hub, subscriberRole, []string{})})
```
Darkweak: Setup constant cookie name with value mercureAuthorization

hub/subscribe_test.go
```diff
@@ -329,6 +333,90 @@ func TestSubscribeTarget(t *testing.T) {
...
+
+			log.Info("Waiting for subscriber...")
+			if ready {
+				break
+			}
+		}
+
+		ctx, cancelRequest2 := context.WithCancel(context.Background())
+		req := httptest.NewRequest("GET", defaultHubURL+"?topic=https://example.com", nil).WithContext(ctx)
+		req.AddCookie(&http.Cookie{Name: "mercureAuthorization", Value: createDummyAuthorizedJWT(hub, subscriberRole, []string{})})
```
Darkweak: Same here

spec/mercure.md
```diff
@@ -324,6 +323,45 @@ after a long disconnection time).
...
+value of the subscribed topic and `subscriptionID` is an unique identifier for this subscription.
+`subscriptionID` **MAY** be a UUID [@RFC4122].
+
+The content of the update **MUST** be a JSON-LD [@!W3C.REC-json-ld-20140116] document containing at
+least the following properties:
+
+ *  `@id`: the identifier of this update, it **MUST** be the same value as the subscription update's
+    topic
+
+ *  `@type`: the fixed value `https://mercure.rocks/Subscription`
```
soyuka: I'm not sure I agree that this should be a fixed value, it should at least be configurable. I mean what happens if you want to change the name, or you get hit by a bus and we can't get the domain?

hub/subscribe_test.go
```diff
@@ -329,6 +333,90 @@ func TestSubscribeTarget(t *testing.T) {
...
+		w := httptest.NewRecorder()
+		hub.SubscribeHandler(w, req)
+
+		resp := w.Result()
+		defer resp.Body.Close()
+		body, _ := ioutil.ReadAll(resp.Body)
+
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		bodyContent := string(body)
+		assert.Contains(t, bodyContent, `data:   "@id": "https://mercure.rocks/subscriptions/https%3A%2F%2Fexample.com/`)
```
Darkweak: Same here

hub/subscribe_test.go
```diff
@@ -329,6 +333,90 @@ func TestSubscribeTarget(t *testing.T) {
...
+		assert.Contains(t, bodyContent, `data:   "@type": "https://mercure.rocks/Subscription",`)
```
Darkweak: Same here

hub/subscribe.go
dunglas: It's not a host name, it's a RDF type. It's similar to Schema.org type: their IDs never change. TL;DR: it's fixed on purpose.

hub/subscribe.go
dunglas: https://github.com/dunglas/mercure/pull/238#discussion_r380677961

It's a fixed value, on purpose so an env var isn't applicable. IMHO using a constant isn't worth it because every value is used only 1 time.

spec/mercure.md
dunglas: See my previous comment. It's not related to hostnames, domains and the like. It's a RDF type. It's similar to a XML namespace. The value never changes, and doesn't have to be dereferencable (the URL may not exist).

hub/subscribe.go
soyuka: okay then.

docs/hub/config.md
```diff
@@ -15,33 +15,34 @@ Most configuration parameters are hot reloaded: changes made to environment vari
...
+| `use_forwarded_headers`          | set to `true` to use the `X-Forwarded-For`, and `X-Real-IP` for the remote (client) IP address, `X-Forwarded-Proto` or `X-Forwarded-Scheme` for the scheme (http or https), `X-Forwarded-Host` for the host and the RFC 7239 `Forwarded` header, which may include both client IPs and schemes. If this option is enabled, the reverse proxy must override or remove these headers or you will be at risk                                                        |
+| `dispatch_subscriptions`         | set to `true` to dispatch updates when a subscription between the Hub and a subscriber is established or closed. The topic follows the template `https://mercure.rocks/subscriptions/{subscriptionID}`. To receive connection updates, subscribers must have `https://mercure.rocks/targets/subscriptions` or an URL matching the template `https://mercure.rocks/targets/subscriptions/{topic}` (`{topic}` is URL-encoded topic of the subscription) as targets |
+| `subscriptions_include_ip`       | set to `true` to include the subscriber's IP in the subscription update
```
jdreesen: I think there's a `|` missing at the end.
```suggestion
| `subscriptions_include_ip`       | set to `true` to include the subscriber's IP in the subscription update                                                                                                                                                                                                                                                                                                                                                                                           |
```

PR #245, opened 2020-03-18
Update install.md

jusquiame:
a closing single quote was missing after '!ChangeMe! in Windows PowerShell launching command


dunglas:
Goood catch! Thanks @jusquiame

PR #260, opened 2020-04-25
 Switch to Github Actions 

dunglas:
Follow up of #196


dunglas:
Thank you very much @jderusse! We finally moved out of Travis.

PR #251, opened 2020-04-09
Hub monitoring

jdecool:
There is actually no way to easily monitor the Hub activity.

This PR adds a way to configure a monitoring system.
I add an implementation for Prometheus with de default metrics provided by the default client.

Metrics are available on the `/metrics` URI.

Output example:

```bash
$ curl http://localhost:3001/metrics

# HELP go_gc_duration_seconds A summary of the GC invocation durations.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 0
go_gc_duration_seconds{quantile="0.25"} 0
go_gc_duration_seconds{quantile="0.5"} 0
go_gc_duration_seconds{quantile="0.75"} 0
go_gc_duration_seconds{quantile="1"} 0
go_gc_duration_seconds_sum 0
go_gc_duration_seconds_count 0
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 10
# HELP go_info Information about the Go environment.
# TYPE go_info gauge
go_info{version="go1.14.1"} 1
# HELP go_memstats_alloc_bytes Number of bytes allocated and still in use.
# TYPE go_memstats_alloc_bytes gauge
go_memstats_alloc_bytes 704392
# HELP go_memstats_alloc_bytes_total Total number of bytes allocated, even if freed.
# TYPE go_memstats_alloc_bytes_total counter
go_memstats_alloc_bytes_total 704392
# HELP go_memstats_buck_hash_sys_bytes Number of bytes used by the profiling bucket hash table.
# TYPE go_memstats_buck_hash_sys_bytes gauge
go_memstats_buck_hash_sys_bytes 4027
# HELP go_memstats_frees_total Total number of frees.
# TYPE go_memstats_frees_total counter
go_memstats_frees_total 255
# HELP go_memstats_gc_cpu_fraction The fraction of this program's available CPU time used by the GC since the program started.
# TYPE go_memstats_gc_cpu_fraction gauge
go_memstats_gc_cpu_fraction 0
# HELP go_memstats_gc_sys_bytes Number of bytes used for garbage collection system metadata.
# TYPE go_memstats_gc_sys_bytes gauge
go_memstats_gc_sys_bytes 3.436808e+06
# HELP go_memstats_heap_alloc_bytes Number of heap bytes allocated and still in use.
# TYPE go_memstats_heap_alloc_bytes gauge
go_memstats_heap_alloc_bytes 704392
[...]

PR #265, opened 2020-04-27
Improve metrics endpoint with Hub data

jdecool:
The PR (#251) adds endpoint to expose system metrics about the Hub.

In this one, I add Hub related metrics:
* `mercure_subcribers` for current number of running subscribers (by topics)
* `mercure_subcribers_total` for total number of handled subscribers (by topics)
* `mercure_updates_total` for total number of handled updates (by topics)

Output:

```text
# ...
# HELP mercure_subcribers The current number of running subscribers
# TYPE mercure_subcribers gauge
mercure_subcribers{topic="http://example.com/alt/1"} 2
mercure_subcribers{topic="http://example.com/foo/1"} 1
# HELP mercure_subcribers_total Total number of handled subsribers
# TYPE mercure_subcribers_total counter
mercure_subcribers_total{topic="http://example.com/alt/1"} 3
mercure_subcribers_total{topic="http://example.com/foo/1"} 1
# HELP mercure_updates_total Total number of handled updates
# TYPE mercure_updates_total counter
mercure_updates_total{topic="http://example.com/alt/1"} 1
mercure_updates_total{topic="http://example.com/foo/1"} 2
# ...
```

I also create a `testServer` struct to simplify testing.


hub/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+			prometheus.CounterOpts{
+				Name: "mercure_updates_total",
+				Help: "Total number of handled updates",
+			},
+			[]string{"topic"},
+		),
+	}
+}
+
+// Register configure the Prometheus registry with all collected metrics.
```
dunglas, line 43:
```suggestion
// Register configures the Prometheus registry with all collected metrics.
```

hub/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+	registry.MustRegister(m.subscribers)
+	registry.MustRegister(m.subscribersTotal)
+	registry.MustRegister(m.updatesTotal)
+
+	// Go-specific metrics about the process (GC stats, goroutines, etc.).
+	registry.MustRegister(prometheus.NewGoCollector())
+	// Go-unrelated process metrics (memory usage, file descriptors, etc.).
+	registry.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
+
+	r.Handle("/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{})).Methods("GET")
```
dunglas, line 57: Shouldn't this endpoint be protected by some authentication mechanism (Basic HTTP?)

hub/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// NewSubscriber collect metrics about new subscriber event.
+func (m *Metrics) NewSubscriber(topics []string) {
```
dunglas, line 61: You should also expose the targets and the IP. Here are the data we currently log: https://github.com/dunglas/mercure/blob/master/hub/log.go#L11-L31

hub/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Go-specific metrics about the process (GC stats, goroutines, etc.).
+	registry.MustRegister(prometheus.NewGoCollector())
+	// Go-unrelated process metrics (memory usage, file descriptors, etc.).
+	registry.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
+
+	r.Handle("/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{})).Methods("GET")
+}
+
+// NewSubscriber collect metrics about new subscriber event.
```
dunglas, line 60:
```suggestion
// NewSubscriber collects metrics about new subscriber events.
```

hub/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+func (m *Metrics) NewSubscriber(topics []string) {
+	for _, t := range topics {
+		m.subscribersTotal.WithLabelValues(t).Inc()
+		m.subscribers.WithLabelValues(t).Inc()
+	}
+}
+
+// SubscriberDisconnect collect metrics about subscriber disconnection event.
```
dunglas, line 68:
```suggestion
// SubscriberDisconnect collects metrics about subscriber disconnection events.
```

hub/publish.go
```diff
@@ -77,6 +71,24 @@ func (h *Hub) PublishHandler(w http.ResponseWriter, r *http.Request) {
 		panic(err)
 	}
 
+	h.metrics.NewUpdate(topics)
```
dunglas, line 74: Can you move that after `io.WriteString(w, u.ID)` and the log (to reduce the latency).

hub/server_test.go
```diff
@@ -234,3 +234,110 @@ func TestMetricsAccess(t *testing.T) {
...
+	s.wgTested.Wait()
+}
+
+func (s *testServer) newSubscriber(topic string, fn func(*http.Response)) {
+	s.wgConnected.Add(1)
+	s.wgTested.Add(1)
+
+	go func() {
+		defer s.wgTested.Done()
+		resp, err := s.client.Get(testURL + "?topic=" + url.QueryEscape(topic)) //nolint:bodyclose
```
dunglas, line 310: the body must be closed here.

hub/subscribe.go
```diff
@@ -116,13 +116,17 @@ func (h *Hub) initSubscription(w http.ResponseWriter, r *http.Request) (*Subscri
...
+	h.metrics.NewSubscriber(topics)
+
 	log.WithFields(fields).Info("New subscriber")
 
 	// Listen to the closing of the http connection via the Request's Context
 	go func() {
 		<-r.Context().Done()
 		pipe.Close()
 
+		h.metrics.SubscriberDisconnect(topics)
```
dunglas, line 128: This will change in #263

hub/metrics.go
jdecool: Not necessarily.

But it should be possible. We could add a parameter to enable authentication, by adding a `METRICS_AUTHENTICATION="user:password"` parameter ?

hub/publish.go
jdecool: :heavy_check_mark:

hub/metrics.go
jdecool: I'm not sure that IP information is relevant in metrics (metrics are not logs).

We could add targets as a label. The output result will me something like:

```text
mercure_subcribers{topic="xxx", target="yyy"}
mercure_subcribers_total{topic="xxx", target="yyy"}
mercure_updates_total{topic="xxx", target="yyy"}
```

But with this, as each label creates a unique combination of time series, it will increase the amount of data.

dunglas: Ok let's keep it as is then!

hub/metrics.go
dunglas: Yes or `METRICS_LOGIN` and `METRICS_PASSWORD`. It can be dangerous to expose such infos over the internet isn't it?

dunglas:
Thanks @jdecool!

hub/metrics.go
jdecool: >  It can be dangerous to expose such infos over the internet isn't it?

Yes it gives information about your infrastructure.

I will make a PR to add authentication on the endpoint.

PR #274, opened 2020-04-30
Replace yaourt with yay in documentation

pocky:
Yaourt is deprecated (https://github.com/archlinuxfr/yaourt) and unmaintained. Yay is (one of) the best alternative since 2019 (https://wiki.archlinux.org/index.php/AUR_helpers#Pacman_wrappers).


dunglas:
Thanks!

Issue #249, opened 2020-04-03
Add version information


Gemorroj:
Add `-v` or `--version` flag to show verion of mercure server.

dunglas:
Good idea, and PR welcome! IMO it should be scripted: we could provide a script that update a Go file containing the version number, create the related Git tag and push it.

jdecool:
This issue can be closed.

#276 is merged.

PR #276, opened 2020-04-30
Add --version flag

jdecool:
Proposal for #249 

Output in development mode :

```text
$ go run main.go --version

Mercure.rocks Hub version dev
https://github.com/dunglas/mercure/releases/latest
```

Output with tagged version : 

```text
$ ./mercure --version

Mercure.rocks Hub version 0.10.0, commit 2159374, build at 2020-04-30T20:11:23Z
https://github.com/dunglas/mercure/releases/tag/v0.10.0
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -44,6 +44,6 @@ func init() { //nolint:gochecknoinits
 	appVersion := common.AppVersion
 	rootCmd.Version = appVersion.Shortline()
 
-	versionTemplate := fmt.Sprintf("mercure version %s\n%s\n", rootCmd.Version, appVersion.ChangelogURL())
+	versionTemplate := fmt.Sprintf("Mercure.rocks Hub version %s\n%s\n", rootCmd.Version, appVersion.ChangelogURL())
 	rootCmd.SetVersionTemplate(versionTemplate)
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,22 +2,14 @@ package cmd
 
 import (
 	"fmt"
-	"regexp"
-	"runtime/debug"
-	"strings"
 
 	log "github.com/sirupsen/logrus"
 
+	"github.com/dunglas/mercure/common"
 	"github.com/dunglas/mercure/hub"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
-
-// version is the running Hub version and is dynamically set at build
-var version = "dev" //nolint:gochecknoglobals
-
-// buildDate stores the build date and is dynamically set at build
-var buildDate = "" //nolint:gochecknoglobals
 
 // rootCmd represents the base command when called without any subcommands.
 var rootCmd = &cobra.Command{ //nolint:gochecknoglobals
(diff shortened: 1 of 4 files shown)


jdecool:
The golangci-lint was updated 2 days ago (v1.26) and the new version seems to break the CI lint job.

dunglas:
Error handling errors fixed by #279. A rebase should make this PR green!

jdecool:
:heavy_check_mark: 

Thanks @dunglas

common/version.go
```diff
@@ -0,0 +1,53 @@
...
+	if v.BuildDate == "" {
+		return v.Version
+	}
+
+	return fmt.Sprintf("%s (%s)", v.Version, v.BuildDate)
+}
+
+func (v *AppVersionInfo) ChangelogURL() string {
+	path := "https://github.com/dunglas/mercure"
+	r := regexp.MustCompile(`^v?\d+\.\d+\.\d+(-[\w.]+)?$`)
```
dunglas, line 31: Maybe can we remove this regex and just check if `version == dev`?

common/version.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"regexp"
+	"runtime/debug"
+	"strings"
+)
+
+type AppVersionInfo struct {
+	Version   string
+	BuildDate string
```
dunglas, line 12: Could we also include de the Git commit? It is provided by GoReleaser.

cmd/root.go
```diff
@@ -37,4 +40,10 @@ func init() { //nolint:gochecknoinits
 	})
 	fs := rootCmd.Flags()
 	hub.SetFlags(fs, v)
+
+	appVersion := common.AppVersion
+	rootCmd.Version = appVersion.Shortline()
+
+	versionTemplate := fmt.Sprintf("mercure version %s\n%s\n", rootCmd.Version, appVersion.ChangelogURL())
```
dunglas, line 47:
```suggestion
	versionTemplate := fmt.Sprintf("Mercure.rocks Hub version %s\n%s\n", rootCmd.Version, appVersion.ChangelogURL())
```

common/version.go
```diff
@@ -0,0 +1,53 @@
+package common
```
dunglas, line 1: Can't we put this code in the `main.go` file? So we'll be able to use GoReleaser's default ldflags: https://goreleaser.com/environment/#using-the-mainversion

common/version.go
jdecool: The regex has been added to deal with the `--snapshot` flag.

This flag can be used to run a build without create a new tag.
In this case the Github link would not be valid.

But it may not be necessary ?

common/version.go
jdecool: :+1: 

Are you OK if I change the output by:

```
Mercure.rocks Hub version 0.10.0, commit 2159374, build at 2020-04-30T20:11:23Z
https://github.com/dunglas/mercure/releases/tag/v0.10.0
```

common/version.go
jdecool: The problem if we moved this code in the `main.go` is that we will have to use a global variable from the `main` package in the `cmd` package. The latter being used to define CLI commands (and manage the output version of the `--version` flag).

Moreover, if this PR is accepted, I've planned to add a metadata metric to expose those information in Prometheus. So this structure will be also used in the `hub` package thought the metrics collector.

That's why, I decided to add the code in a `common` package.

common/version.go
dunglas: perfect

common/version.go
dunglas: Ok let's keep this as is then!

dunglas:
Thanks @jdecool

PR #280, opened 2020-05-07
Fixed a typo in openapi.yaml

lyderichti59:
Fixed a typo in openapi.yaml's description of the get endpoint for subscribing to updates


dunglas:
Good catch!

PR #288, opened 2020-05-12
Major spec changes

dunglas:
This a major spec change intended to fix most remaining limitations. I hope that it can be the last breaking change.

# Authorization

This is the main change.

The authorization mechanism has been simplified. It's currently the major pain point for users. The concept of `target` is difficult to understand. In this proposal, targets are gone. The new mechanism is much simpler:

* An update can be marked as `private` by setting the `private` property while publishing it.
* In order to receive private updates, subscribers must have a JWT, and a URI template matching the private update's topic must be present in the `mercure.subscribe` claim of this JWT. It's also possible to use a raw string (it can be a URL or just a random identifier) instead of a URI template.

# Subscription events

* The topic of the subscription event now follow the pattern `/.well-known/mercure/subscriptions/{subscriptionID}/{topic}` (it was `htts://mercure.rocks/subscriptions/{topic}/{subscriptionID}`). Using well-known URL for this looks cleaner to me.
* As the target mechanism is gone, events are now private updates.
* It is now possible to store user-defined data in the JWT of the subscriber (in `mercure.payload`), and this user-defined payload is included in the subscription events. It is convenient to store a user ID, an IP, groups and so on.

# Event Store and State Reconciliation

The Hub can now be used as an event store (to implement event sourcing for instance) thanks to a simple trick: the event ID `-1` is now reserved, and using `-1` as value for the `Last-Event-ID` header or query parameter now hints the hub to send the full history for the subscribed topic(s).

I'm not sure about `-1`, it could be `0` too, or `all`,  or any other reserved value. Don't hesitate to share your thoughts about this.

# Identifiers

It's now possible and recommended to use IRIs everywhere: as topics and as event ids. For convenience, it's also possible to use random string (not IRIs) everywhere.

# Encryption

The key set (JWKS) is now exposed through an external URL.

# Next steps

* The hub must be updated accordingly.
* I also plan to define a tiny REST API in the spec allowing to fetch the list of active subscriptions.

### DIFF
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -38,21 +38,20 @@ The keywords **MUST**, **MUST NOT**, **REQUIRED**, **SHALL**, **SHALL NOT**, **S
 NOT**, **RECOMMENDED**, **MAY**, and **OPTIONAL**, when they appear in this document, are to be
 interpreted as described in [@!RFC2119].
 
- *  Topic: The unit to which one can subscribe to changes. The topic **MUST** be identified by an
-    IRI [@!RFC3987] or by a string. Using an HTTPS [@!RFC7230] or HTTP [@!RFC7230] URI [@!RFC3986]
-    is **RECOMMENDED**.
+ *  Topic: The unit to which one can subscribe to changes. The topic **SHOULD** be identified
+    by an IRI [@!RFC3987]. Using an HTTPS [@!RFC7230] or HTTP [@!RFC7230] URI [@!RFC3986] is
+    **RECOMMENDED**.
+
+ *  Update: The message containing the updated version of the topic. An update can be marked as
+    private, consequently, it must be dispatched only to subscribers allowed to receive it.
 
  *  Publisher: An owner of a topic. Notifies the hub when the topic feed has been updated. As in
     almost all pubsub systems, the publisher is unaware of the subscribers, if any. Other pubsub
     systems might call the publisher the "source". Typically a website or a web API, but can also be
     a web browser.
 
- *  Subscriber: A client application that subscribes to real-time updates of topics. Typically a
-    Progressive Web App or a Mobile App, but can also be a server.
-
- *  Target: A subscriber, or a group of subscribers. A publisher is able to securely dispatch
-    updates to specific targets. The target **MUST** be identified by an IRI [@!RFC3987] or by a
-    string. Using an HTTPS [@!RFC7230] or HTTP [@!RFC7230] URI is **RECOMMENDED**.
+ *  Subscriber: A client application that subscribes to real-time updates of topics. Typically a web
+    or a mobile application, but can also be a server.
 
  *  Hub: A server that handles subscription requests and distributes the content to subscribers when
     the corresponding topics have been updated. Any hub **MAY** implement its own policies on who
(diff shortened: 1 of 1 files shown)


dmouse:
About mercure.payload, can it be used as a sub-topic?

For example: 
This messages is delivered only for subscribed galgo dogs

htts://mercure.rocks/subscriptions/vendor/dogs
{
  breed: "galgo", 
}

Is it correct?

dunglas:
@dmouse Sorry I don't understand exactly what you mean. The payload is a random data structure you add to the subscriber's JWT. Then, this data structure is attached to the subscriber, and included in subscription events related for this subscriber.

spec/mercure.md
```diff
@@ -323,127 +353,145 @@ allowed to, a 403 HTTP status code **SHOULD** be returned.
...
 If both the `Last-Event-ID` HTTP header and the query parameter are present, the HTTP header
 **MUST** take precedence.
 
 If the `Last-Event-ID` HTTP header or query parameter exists, the hub **SHOULD** send all events
 published following the one bearing this identifier to the subscriber.
 
-The hub **MAY** discard some messages for operational reasons. The subscriber **MUST NOT** assume
-that no update will be lost, and **MUST** re-fetch the original topic to ensure this (for instance,
-after a long disconnection time).
+The reserved value `-1` can be used to hint the hub to send all updates it has for the subscribed
```
sroze, line 428: This is confusing to me. I'd have expected `-1` to actually be used as a relative value to only return the last event. I'd go for a much more explicit `all` constant.

a-menshchikov:
I believe that the rejection of the mechanism of targets was the wrong decision.

Suppose we have a single page application that contains two thousand entries. Even entries owned by userA, and odd entries owned by userB. An entry have properties "id", "status" and "owner". (It's important that an application is SPA: user get an authorization token for Mercure only once, when start an application.)
I want to subscribe users for status updates of their entries. *But without the targets mechanism it seems impossible*: user **must** subscribe for concrete topic if we want to make it private (e.g. "https://example.com/entry/123/status"), but if user own a thousand of entries authorization token will have extremely large size and it will be unusable both in cookie or in authorization header  while connecting to Mercure.

@dunglas could you describe an example of solution for my example case, when we need to publish status updates in SPA for users with a lot of entries and we can't subscribe it for concrete entries by their enumeration?

dunglas:
@a-menshchikov you can use topic selectors for this use case: `https://example.com/entry/{id}/status`. If you want user A to receive some updates, and user B receive some others, you can leverage the alternate topic feature:

    POST /.well-known/mercure

    topic=https://example.com/entry/1/status&topic=https://example.com/entry/forUserA/1/status&private=true

    POST /.well-known/mercure

    topic=https://example.com/entry/2/status&topic=https://example.com/entry/forUserB/2/status&private=true

Then you can use `https://example.com/entry/forUserA/{id}/status` as target selector for user A, and `https://example.com/entry/forUserB/{id}/status` for user A.

What do you think?

a-menshchikov:
@dunglas it seems less clear than it was with targets mechanism, especially if we have tens of different entry classes in our application. Previously, we had to generate JWT for user with one target for himself (`user/{username}`) and set appropriate target on publication. Now we should generate JWT with enumeration of all kind of entries in topic.

As result we have more complicated code of JWT generation and we cannot segregate JWT generator and core application because it should know about entry classes.

a-menshchikov:
Also this approach requires me to solve an issue how to pass the username to client (that need for a subscription url entry/**forUserA**/{id}/status).

With targets mechanism client doesn't need to know anything about private subscription principle. It basically get JWT from backend that contains information about allowed targets, and then retrieve from hub granted messages. It's much more simpler than need to make JWT with all private topics and pass the username to client.
Finally, it seems extremely hard to implement subscription simultaneously for certain users and user groups.

dunglas:
> Previously, we had to generate JWT for user with one target for himself (user/{username}) and set appropriate target on publication. Now we should generate JWT with enumeration of all kind of entries in topic.

This can be solved using URI templates.
The JWT can contain something like `mercure.subscribe: ["https://example.com/userA{/any*}"]`. Then you can use alternate IRIs just as you were using targets before:

```
POST /.well-known/mercure

topic=https://example.com/foo/123&topic=https://example.com/userA/foo/123
```

> Also this approach requires me to solve an issue how to pass the username to client

No because it's just an alternate IRI. Client can still refer to the canonical IRI which doesn't need to contain a reference to the client name.

Regarding the previous example, the client can still subscribe to `https://example.com/foo/123` or even `https://example.com/foo/{id}`.

a-menshchikov:
@dunglas this solution works, thank you!

bats-dm:
@dunglas So if I want to publish the same update for the several users I need to use multiple topics, right?
```
POST /.well-known/mercure

topic=https://example.com/userA/foo/123&topic=https://example.com/userB/foo/123&topic=https://example.com/user{N}/foo/123&private=on
```
For example, we have the chatroom in our chat. This chatroom contains 1000 users at the moment. Will it be 1000 topics in POST request with just different users?

dunglas:
For a chat room there is an easier way:

* messages can have a topic following a pattern similar to `https://example.net/the-chat-room/<id>` (example: POST topic=`https://example.net/thé-chat-room/22`)
* users can bear a JWT containing the template selector `https://example.net/thé-chat-room/{id}`

This way, you every messages can have a single topic, and all authorized users can receive them.

bats-dm:
@dunglas I thought about this way, but if we add `https://example.net/the-chat-room/{id}` to JWT all users have access to all chat rooms and can read messages. But for security reasons they should have access only for rooms they are in. The only way I see is to update the user's JWT every time they in or out a chat room. But the problem that only admins can assign or unassign the user to the room.

 So I consider the way I mentioned in the previous post but I want to make sure that I understand everything correctly because I used targets earlier and I want to update Mercure.

Issue #261, opened 2020-04-26
Looking for a solution to fetch user info from a new connection


josephmiller2000:
We are building a chat application using symfony and mercure.

We have achieved a lot of functions and all going smooth until we tried to get user info.

### What we are trying to achieve here:

- We are creating a massive user list at the right side of the chat room, where chatters can see the list of active chatters in the room.

- We used "Ping" method to fetch total user information redis storage and list it them on the user list.

-  This method is ok for a small chat with couple 10's of chatters. Imagine what happens when we get 1000+ users and the list keep growing more and more? each ping should receive more than 1000+ user info in each ping. This is not a right way to do it.

- So we are planning to receive each user info ( name, and user account details ) from the new connected users and add them (append them) to the user list. ==> One by One to the user list.

- So now lets go to the disconnect function, when a user logs out, or disconnected for some reason. The name of the user should be removed from the user list.

This is what we are trying to achieve, and now we planned to use websocket for this, but before using websocket, i want to give a last final try to use mercure or any other suggestive method to achieve this

Our current project demo link >> https://symfony.tamil2friends.com/guest

Any open suggestion are welcome, and we want to do make it right way and very efficient way.

dunglas:
Did you try to use the new [subscription events](https://mercure.rocks/spec#subscription-events)? 

You can use these new events to notify all currently connected users when a new user is joining, or when a user is leaving (and update the list accordingly).
To be pass the username associated to this event to other users, be sure to include it as a [target](https://mercure.rocks/spec#terminology). I suggest to add a target such as `https://chat.example.com/user/<the-username>` in the JWT of every user, so you can easily distinguish this ID containing the username from other targets.

With this approach, there is still one remaining problem: when a new user joins, it must be able to retrieve the list of all currently connected users. What I suggest to do is to store this list in your redis or in your database. Basically, a server component must register to the subscription events, and update the list every time a user connects or disconnects. Then, you can expose this list through an API endpoint in your Symfony app.

To summarize:

**config**
1. Enable the subscription events in the Mercure Hub configuration
2. Include the username (or ID) in as a target in the Mercure JWT, also include the `https://mercure.rocks/targets/subscriptions/<your-chat-topic>` target to allow users to receive subscription/unsubscription events 

**server-side**
1. Create a script that subscribes to this event (if you want to use PHP, you can use a SSE client such as https://github.com/clue/reactphp-eventsource). This script will store the list of currently connected users in Redis, or in database.

[...]

Issue #246, opened 2020-03-19
mercure hub subscriber list 


benCalmette:
We arer trying to list all subscribers on a topic, unfortunately we can't use subsribtion url's or any solution to achieve this. Maybe with an access to bolt database from outside the docker container.
Someone can tell me what is subscriptionID ?

@see last comment on #62

dunglas:
Currently, such endpoint is not available. At this point, the hub just publishes events in a specific topic when a subscriber connects or disconnects. It doesn't make the list of connected users accessible yet. The reason is that when you use a custom transport or the HA version, it will be more than one Mercure Hub, and subscribers must be aggregated across all hubs. 

One current workaround is to store this state externally (for instance in your API's database) by subscribing to these connection events and updating your DB when one of them occur.

We plan to add an endpoints to list currently connected users at some point, but it needs more work. We also plan to add support for the WebSub protocol (#100), so it will be easier for servers to subscribe to events (by receiving POST requests instead of using EventSource).

PR #295, opened 2020-05-18
Spec: add a web API to retrieve active subscriptions

### LINKED ISSUE #117: A way to know how many clients are connected to the hub ?

dunglas:
Close #117, #246, #261.

This PR introduces a tiny REST API which can be used to retrieve the list of active connections. This is useful for new subscribers to retrieve the current list of connected users  (ex: chat room). This API can also be useful for servers (publishers) to know if a user is currently connected or not.

### DIFF
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -468,12 +468,15 @@ If a subscriber has several subscriptions, it **MUST** be identified by the same
 The content of the update **MUST** be a JSON-LD [@!W3C.REC-json-ld-20140116] document containing at
 least the following properties:
 
- *  `@id`: the identifier of this update, it **MUST** be the same value as the subscription update's
+ *  `@context`: the fixed value `https://mercure.rocks/`. `@context` can be omitted if already
+    defined in a parent node. See (#json-ld-context).
+
+ *  `id`: the identifier of this update, it **MUST** be the same value as the subscription update's
     topic
 
- *  `@type`: the fixed value `https://mercure.rocks/Subscription`
+ *  `type`: the fixed value `Subscription`
 
- *  `topicSelector`: the topic selector used of this subscription
+ *  `topic`: the topic selector used of this subscription
 
  *  `subscriber`: the topic identifier of the subscriber. It **SHOULD** be an IRI.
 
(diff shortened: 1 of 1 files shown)


spec/mercure.md
```diff
@@ -479,6 +487,169 @@ The JSON-LD document **MAY** contain other properties.
...
+   "topicSelector": "https://example.com/{selector}",
+   "subscriber": "urn:uuid:bb3de268-05b0-4c65-b44e-8f9acefc29d6",
+   "active": true,
+   "payload": {"foo": "bar"}
+}
+~~~
+
+# Subscription API
+
+If the hub supports subscription events (see (#subscription-events)), it **MUST** also expose active
```
docteurklein, line 505: Is it a good idea to force implementors to maintain a wab API if they're not interested in "presence" ?
Shouldn't that be optional?

dunglas: It's optional. What is mandatory is to expose this API if subscription events are supported. I'm not sure that it makes sense to have subscription events without a way to retrieve the current active connections. At least, I'm not able to find a use case for this. But I may be wrong. WDYT?

spec/mercure.md
```diff
@@ -447,29 +450,34 @@ response, however polyfills and server-sent events clients in most programming l
...
 
- *  `@id`: the identifier of this update, it **MUST** be the same value as the main subscription
-    update's topic
+ *  `@id`: the identifier of this update, it **MUST** be the same value as the subscription update's
+    topic
 
  *  `@type`: the fixed value `https://mercure.rocks/Subscription`
 
- *  `topic`: the topic to which the subscription refers
+ *  `topicSelector`: the topic selector used of this subscription
```
dunglas, line 476: Should we use `topic` as the query parameter, or should we rename the `topic` query parameter to `topicSelector`? (Ex: `https://example.com/.well-known/mercure?topicSelector=https://example.com/foo/{bar}`)

dunglas:
Closed by #295 and #298.

spec/mercure.md
dunglas: I changed my mind, see #311.

PR #297, opened 2020-05-20
Use "earliest" instead of "-1" as reserved value to retrieve all events

dunglas:
See https://twitter.com/dunglas/status/1262675844352409604


sroze:
Beautiful.

PR #352, opened 2020-07-14
Try to clarify difference between protocol and hub

wouterj:
There was quite some confusion in the Symfony Slack about Mercure yesterday. I was completely confused as well and I'm happy things are clarified now, thank you for your time @dunglas! To give something back for the great OSS protocol and hub, I wanted to see whether the docs can be more clear to hopefully avoid some confusion.

This PR changes 3 things. I'm happy to revert things if you disagree, we're just about getting started with Mercure.

1. Explicitly call the hub "Mercure.rocks hub" instead of "Mercure", "Mercure Hub", or just "Hub". To be more explicit about what is talking about the hub in a generic (protocol-based) sense and what is talking about the Go hub;
2. Changed the wordings in the "Starting the Hub" section of the Getting Started guide. This is probably the most read guide for beginners, it's important to clarify the different entities (Protocol, Go hub, etc) and their meaning;
3. Removed "Upgrade to newest versions" from the Protocol Specification. It seems a changelog for the hub (when looking at the first sentence of this doc).

I recommend some other changes, which don't live in this repository:

* The purchasing information about the on-premise HA version is hidden [far deep in the docs](https://mercure.rocks/docs/hub/cluster#purchasing). I recommend adding this info on the "Pricing" page as well. I only found this information while working on this PR (and this is the option I'm interested in, so I had been looking for it).
* The AGPL license is in a list of licenses that most people are afraid of (I have to discuss it with my COO before I'm allowed to use AGPL software). I recommend adding something about the implications of this license in the docs. As I understand, you should not be afraid of the AGPL license if you use the Hub, but that's only after you posted it on Slack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ The protocol has been published as [an Internet Draft](https://datatracker.ietf.
 A reference, production-grade, implementation of [**a Mercure hub**](https://mercure.rocks/docs/hub/install) (the server) is also available in this repository.
 It's a free software (AGPL) written in Go. It is provided along with a library that can be used in any Go application to implement the Mercure protocol directly (without a hub) and an official Docker image.
 
-In addition, a managed and high-scalability version of Mercure is [available on Mercure.rocks](https://mercure.rocks/pricing).
+In addition, a managed and high-scalability version of the Mercure.rocks hub is [available on Mercure.rocks](https://mercure.rocks/pricing).
 
 ## Contributing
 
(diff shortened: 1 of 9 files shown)


dunglas:
Thanks for taking the time of clarifying @wouterj! I'll also make the changes you suggested to the website.
Regarding AGPL, I could add the statement as for the Vulcain Gateway Server: https://github.com/dunglas/vulcain#license-and-copyright

WDYT?

wouterj:
> Regarding AGPL, I could add the statement as for the Vulcain Gateway Server: https://github.com/dunglas/vulcain#license-and-copyright
>
> WDYT?

That seems very clear and concise :+1:

Thanks for the quick merge

Issue #357, opened 2020-07-27
Move the Helm chart to this repo


sagikazarmark:
As of Nov 2020 the "official" (central) helm chart repository will be marked obsolete. The best practice now is to host charts next to the project (either in the same repo or in a dedicated chart repo).

It also makes easier to merge contributions/keep changes in a single PR.

The chart repository itself can be hosted in a github repo, using github pages and a custom domain (eg. `charts.mercure.rocks`)

I'd also drop Helm 2 support and focus on Helm 3.

The chart in the official Helm repository can be kept for a period of time so that people have time to transition to the new chart.

I'd be happy to provide a PR if you are interested.

dunglas:
Hi @sagikazarmark, thanks for the detailed report. A PR will be very appreciated!

sagikazarmark:
Sure. What are your options for running a Helm chart repository? It turned out the github repository as a helm repository thing is pretty dumb, I'm working on a github plugin now. Alternatively, you could setup a GCS or S3 bucket, or a ChartMuseum instance, but that last one is probably an overkill.

dunglas:
I'll probably go for Chart Museum as I have a K8s cluster available for Mercure. I'll take a look.

sagikazarmark:
Still valid

sagikazarmark:
Still valid

asimonf:
Hey guys, any chance we can get this up somewhere? It has been a pain to get a chart for this and I'm currently running the deprecated version. I think there's interest to help get this up, but without guidance on hosting this will probably remain unresolved.

dunglas:
We'll host it directly in this repository. I've started something here: https://github.com/api-platform/api-platform/tree/main/helm/api-platform

Basically, what is left to do is to extract the Mercure part.

sagikazarmark:
I'm working on a generic Caddy helm chart. It should work with mercure as well.

https://artifacthub.io/packages/helm/sagikazarmark/caddy

Issue #242, opened 2020-02-27
[enhancement] Prometheus metrics endpoint


dvc:
It is strongly needed in production to monitor service state and statistics.

Health check is minimum.

But more stats is better :) 

For example:

- service version, 
- start time (to calculate uptime), 
- memory
- cpu
- disk_size/history_size/history_max
- connections count, 
- messages rate, 
- errors rate, 
- pub/sub auth errors rate
- ...

Any format is good. May be Prometheus? :)

dunglas:
👍 for Prometheus!

jdecool:
Some metrics has been added by #251, #265, #276 & #287

MinDBreaK:
Would it be possible to disable authentication on the metrics endpoint ? 

When you are using annotation-based scraping in Kubernetes, you cannot set basic auth, and you have to set it globally in Prometheus, which would probably conflict if you have some services that check authorization even if not mandatory. Or if multiple teams use the same cluster.

Maybe a source check or a dedicated endpoint could be used, like Traefik does ?

jdecool:
> Would it be possible to disable authentication on the metrics endpoint ?

Authentication on metrics endpoint is not mandatory. If you don't set a login or password the authentication is not enabled.

Isn't that enough ?

If you disable the authentication on metrics endpoint, you could set the security configuration in Kubernetes ?

MinDBreaK:
Actually I though because of the configuration of the Helm Chart that generate credentials, so I though It was mandatory.

Yes, I could set the security configuration from kubernetes by disabling the access to the path. 
However, I think a dedicated port could be a nice to ease the configuration. But that would only be a nice to have. 

Thanks for the answer, I'll just for now update the helm chart !

sagikazarmark:
I agree, exposing the metrics endpoint on a different port is a best practice in the kubernetes world. Then the whole authentication thing becomes unnecessary in the first place.

dunglas:
Pull request welcome! But please keep authentication support, it can be helpful outside of the k8s world.

sagikazarmark:
Can we make that explicit (eg. an enabled/disabled flag)? Implicitly enabling features is generally not a good idea IMHO. The default generated values of the Helm chart is a great example why.

sagikazarmark:
Bump, there is a PR.

PR #358, opened 2020-07-28
Improve telemetry

sagikazarmark:
This PR exposes the telemetry endpoint under a new address (as suggested in #242). It's supposed to deprecate the previous "metrics" configuration and endpoints.

In the kubernetes world (but IMHO in other places too) telemetry information should not be exposed to the whole world. Mercure currently provides authentication support, but exposing telemetry on a separate port, closed from public access is IMHO better.

This is a first version demonstrating the feature. It lacks the authentication support requested by @dunglas in https://github.com/dunglas/mercure/issues/242#issuecomment-664368534, but since it's on a different port, I doubt builtin authentication will actually be useful. I'd probably suggest in the documentation to setup a proxy with authentication if it's absolutely necessary. Putting credentials in a config file is not secure anyway. That being said, I can add authentication support if you still think it's necessary.

Ideally, the "old" metrics endpoint should be deprecated and removed at some point. If you are not concerned with backwards compatibility, I can remove it in this PR (or a separate one). I would also make telemetry enabled by default.

I also added the healthz endpoint to the telemetry address. I would use this endpoint as the liveness check in Kubernetes.

This PR also requires some changes in the [Helm chart](https://github.com/helm/charts/tree/master/stable/mercure), but before adding changes there, I'd love to discuss #357.


dunglas:
> If you are not concerned with backwards compatibility, I can remove it in this PR (or a separate one). I would also make telemetry enabled by default.

BC isn't a concern yet, so don't hesitate to these changes in this PR.
Your rationale about the authentication makes sense, so let's keep it as is.

hub/config.go
```diff
@@ -32,6 +32,9 @@ func SetConfigDefaults(v *viper.Viper) {
 	v.SetDefault("metrics", false)
 	v.SetDefault("metrics_login", "mercure")
 	v.SetDefault("metrics_password", "")
+
+	v.SetDefault("telemetry.enabled", false)
```
dunglas, line 36: We don't use dot in keys yet (but a underscore). Also can we keep "metrics"? telemetry refers to another thing to me (the ability for the editor to gather stats, and we currently don't do this).

sagikazarmark: I chose telemetry, because then it's extensible in the future with things like distributed tracing. Also, I find dot notated config (nested yaml). I can change it, if it doesn't make sense or consistency has a priority now.

dunglas: Yes please change it. Consistency as priority now and I fear the dot will not work with other supported config formats.

sagikazarmark:
@dunglas just to clarify before making any changes (this PR mentions quite a few):

- remove the old metrics endpoint, config and authentication (and update the documentation accordingly)
- rename the telemetry config back to metrics and remove the dot
- make metrics enabled by default

Does that make sense?

dunglas:
It makes sense, but I wouldn't enable it by default.

hub/config.go
```diff
@@ -32,6 +32,9 @@ func SetConfigDefaults(v *viper.Viper) {
...
+	v.SetDefault("telemetry.addr", "127.0.0.1:3002")
```
dunglas, line 37: Shouldn't we register a dedicated default port for Mercure here: https://github.com/prometheus/prometheus/wiki/Default-port-allocations ?

sagikazarmark:
@dunglas implemented the requested changes.

As for registering a dedicated port: it doesn't really matter IMHO, but could be an option outside of this PR.

sagikazarmark:
This test failure is very interesting: if I run that test only, it succeeds. If I run tests for the whole package, it fails.

sagikazarmark:
@dunglas I finally found the problem causing the tests to fail and fixed it. I think Mercure would really benefit from a rewrite as you pointed out in https://github.com/dunglas/mercure/issues/366#issuecomment-689100606. IMHO more wiring should happen in `main.go` and less inside the `Hub` structure.

dunglas:
I agree, I'm actually doing it on Vulcain (another projet I maintain, which more ore less share the same structure than Mercure), and I'm pretty happy with the result: https://github.com/dunglas/vulcain/pull/60

I plan to do the same kind of changes to Mercure.

sagikazarmark:
Sounds cool! Looking forward to it in Mercure. Let me know if I can help. In the meantime: what should we do with this PR?

You could merge it before the refactor, but in case you'd prefer improving it after the refactor, then I guess it would make sense to do it from scratch again. WDYT?

dunglas:
I think we should merge it ASAP. I've no ETA yet for the refactor.

sagikazarmark:
Is there anything else you need from me to merge it?

hub/config.go
```diff
@@ -84,9 +85,11 @@ func SetFlags(fs *pflag.FlagSet, v *viper.Viper) {
 	fs.BoolP("demo", "D", false, "enable the demo mode")
 	fs.StringP("log-format", "l", "", "the log format (JSON, FLUENTD or TEXT)")
 	fs.BoolP("subscriptions", "s", false, "dispatch updates when subscriptions are created or terminated")
-	fs.BoolP("metrics", "m", false, "enable metrics")
-	fs.StringP("metrics_login", "", "mercure", "the user login allowed to access metrics")
-	fs.StringP("metrics_password", "", "", "the user password allowed to access metrics")
+
+	fs.Bool("metrics-enabled", false, "enable metrics")
+	_ = v.BindPFlag("metrics_enabled", fs.Lookup("metrics-enabled"))
```
dunglas, line 90: Why are these calls to `BindPFlag` necessary? It's done automatically line 94.

hub/server.go
```diff
@@ -183,47 +195,31 @@ func (h *Hub) baseHandler(acmeHosts []string) http.Handler {
 	mainRouter.UseEncodedPath()
 	mainRouter.SkipClean(true)
 
-	// Register /healthz and /metrics (if enabledà in way that doesn't pollute the HTTP logs.
+	// Register /healthz and /metrics (if enabled in way that doesn't pollute the HTTP logs).
```
dunglas, line 198:
```suggestion
	// Register /healthz (if enabled, in a way that doesn't pollute the HTTP logs).
```

hub/server.go
```diff
@@ -183,47 +195,31 @@ func (h *Hub) baseHandler(acmeHosts []string) http.Handler {
...
 	handler := h.chainHandlers(acmeHosts)
 	mainRouter.PathPrefix("/").Handler(handler)
 
 	return mainRouter
 }
 
+func (h *Hub) metricsHandler() http.Handler {
+	router := mux.NewRouter()
+
+	router.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
dunglas, line 212: Maybe could we create a new `registerHealthz`function doing this, to factorize both `/healthz` handlers.

hub/server_test.go
```diff
@@ -390,11 +349,11 @@ func TestMetricsCollect(t *testing.T) {
 
 func TestMetricsVersionIsAccessible(t *testing.T) {
 	v := viper.New()
-	v.Set("metrics", true)
+	v.Set("metrics_enabled", true)
 	server := newTestServer(t, v)
 	defer server.shutdown()
 
-	resp, err := server.client.Get("http://" + testAddr + "/metrics") //nolint:noctx
+	resp, err := server.client.Get("http://" + testMetricsAddr + "/metrics") //nolint:noctx
```
dunglas, line 356: You could also check if `/healthz` can be accessed here.

hub/server.go
```diff
@@ -29,6 +29,18 @@ func (h *Hub) Serve() {
...
+	if h.config.GetBool("metrics_enabled") {
+		addr := h.config.GetString("metrics_addr")
+
+		h.metricsServer = &http.Server{
+			Addr:    addr,
+			Handler: h.metricsHandler(),
+		}
+
+		log.WithFields(log.Fields{"addr": addr}).Info("Mercure metrics started")
+		go h.metricsServer.ListenAndServe()
```
dunglas, line 41: This server must be shutdown too in `listenShutdown()` is started.

dunglas (changes requested):

hub/config.go
sagikazarmark: They are not, leftovers from the dot notated version. Done

hub/server.go
sagikazarmark: Done

hub/server.go
sagikazarmark: Done.

hub/server_test.go
sagikazarmark: Done

hub/server.go
sagikazarmark: Done

sagikazarmark:
@dunglas updated the PR with the requested changes.

dunglas (approved):

dunglas:
Thank you very much @sagikazarmark!

hub/config.go
dunglas: Shouldn't we change the port to one we'll add to this list @sagikazarmark?

dunglas: 9755 for instance

sagikazarmark: It doesn't matter on Kubernetes, you can change it from config everywhere else. IMHO it doesn't matter much what the default port is, it's up to you.

PR #360, opened 2020-08-18
Fix metrics typo

Te4g:


dunglas:
Good catch! Could you update the tests accordingly (`TestMetricsCollect`)?

Te4g:
> Good catch! Could you update the tests accordingly (`TestMetricsCollect`)?

Yep, done! Thanks :)

dunglas:
Thanks @Te4g

PR #365, opened 2020-09-02
Document docker-compose setup

paulandrieux:


dunglas:
Thanks @paulandrieux!

PR #374, opened 2020-09-16
Test : subscriber auth for load tests

rneuter:
This PR goal is to add JWT authentication support for **subscribers** in load tests.
I had the need to test a Mercure infrastructure which support authenticated subscribers only.

I am not very happy with [this diff](https://github.com/dunglas/mercure/compare/master...rneuter:test/subscriberAuthForLoadTests?expand=1#diff-07d4e3a1d6f87c8dd0a9caeb2a995444R66) because it implies that Authorization is always send, but empty when **SUBSCRIBER_JWT** environment variable is not setup. What do you think about it ? I had struggle to condition it, 1st time reading and writing Scala...

Thank you for considering these modifications.

### DIFF
--- a/docs/hub/load-test.md
+++ b/docs/hub/load-test.md
@@ -17,6 +17,7 @@ Available environment variables (all are optional):
 
  * `HUB_URL`: the URL of the hub to test
  * `JWT`: the JWT to use for authenticating the publisher
+ * `SUBSCRIBER_JWT`: the JWT to use for authenticating the subsribers
  * `INITIAL_SUBSCRIBERS`: the number of concurrent subscribers initially connected
  * `SUBSCRIBERS_RATE_FROM`: minimum rate (per second) of additional subscribers to connect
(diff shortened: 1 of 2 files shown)


dunglas:
Hi, thanks for this contribution, I would be glad to merge it. However this will break the load test when the env var will not be provided because `Authorization` will not be empty, it will contain `Bearer `.

You can probably store the argument passed to exec in a temporary variable to build it and add the header conditionally if the env var is provided?

rneuter:
Hi,

After reading some Scala and Gatling documentations, I gave it another shot this morning. I had to define a new `subscriberTest` method to add the header conditionally. Works great in both cases (variable present or not). Branch history has been cleaned, I let you check this PR.

Thank you.

dunglas (approved):

dunglas:
Thanks @rneuter! Your contribution is very appreciated!

PR #380, opened 2020-10-06
Add fallback to namespaced claim

### LINKED ISSUE #343: Add namespacing support on claims
Hello,

The aim being to allow the use of Mercure with access tokens generated by Auth0.

https://auth0.com/docs/tokens/guides/create-namespaced-custom-claims
_By default, Auth0 always enforces namespacing; any custom claims with non-namespaced identifiers will be silently excluded from tokens._

Here is an example of Auth0 rule :

```
function` (user, context, callback) {
 var namespace = 'https://example.com/';
 context.accessToken[namespace + 'mercure'] = { "subscribe": [ user.email ] };
 return callback(null, user, context);
}
```

Thanks for your feedback guys and cheers
cc @dunglas

TamasSzigeti:
Support jwt claim with `https://mercure.rocks/` key along with `mercure` for compatibility with Auth0 that strips non-listed claims

resolves #343


dunglas (approved):
Awesome, thanks!

dunglas:
Thanks @TamasSzigeti!

PR #383, opened 2020-10-14
doc: fix curl example

Remy-Luciani:
Hello, it seems the curl example is not working without specifying the name of the header carrying the JWT. I lost a lot of time but learned a lot on Mercure. :) I don't know if it's the right process to suggest this change but I'm really late on my project, sorry! Cheers.


soyuka (approved):
good catch

dunglas:
Good catch indeed! Thank you very much.

PR #386, opened 2020-10-28
docs: mention setting the JWT algo when using public keys

dkarlovi:


dkarlovi:
Eh, the CI setup seems broken, it's failing for an unmodified file.

dunglas:
Thanks @dkarlovi. I'll take care of the CI (failure are intended, we upgrade the linter automatically).

PR #393, opened 2020-11-18
feat!: Caddy module and config refactoring

dunglas:
This is the next step to make Mercure usable as a standalone library and to use Caddy instead of our own HTTP server by default.

Follows #389. Cc @sagikazarmark.


hub/hub.go
```diff
@@ -4,80 +4,198 @@ import (
...
+	return func(h *Hub) {
+		h.transport = t
 	}
 }
 
-// Start is an helper method to start the Mercure Hub.
-func Start() {
-	h, err := NewHub(viper.GetViper())
+// WithTransportURL sets the transport to use by parsing the provided URL.
+func WithTransportURL(tu string, l Logger) Option {
```
sagikazarmark, line 112: I think the Hub should accept a transport directly. When using it as a library, the transport URL might have a different form.

dunglas: I was working on this while you were reviewing, sorry. Could you check the latest version? I totally refactored this part.

sagikazarmark: Sure!

hub/hub.go
```diff
@@ -4,80 +4,186 @@ import (
...
-	logger             Logger
-	transport          Transport
-	server             *http.Server
-	metricsServer      *http.Server
-	topicSelectorStore *TopicSelectorStore
-	metrics            *Metrics
+const HS256 = "HS256"
+
+// Option instances allow to configure the library.
+type Option func(h *Hub)
```
sagikazarmark, line 18: This is a little bit problematic. Someone could easily modify the `Hub` struct after initialization:

```go
hub := New()

hub.Start()

WithDebug()(hub)
```

This can lead to unintended consequences. (This is long overdue to be fixed in my post)

One trick you can do is this:

```go
// Option instances allow to configure the library.
type Option option

type option func(h *Hub)
```

This way the API is not public, only the types are.

The other thing you can do is moving all configuration into an unexported struct:

```go
// Option instances allow to configure the library.
type Option func(o *options)

type options struct {
    debug bool
}
```

It allows you to somewhat separate the configuration itself from the main Hub. For example you could have a `newOptions` function in a separate file that creates a new options struct with defaults.

hub/hub.go
```diff
@@ -4,80 +4,186 @@ import (
...
+	return func(h *Hub) {
+		h.publishOrigins = origins
 	}
 }
 
-// Start is an helper method to start the Mercure Hub.
-func Start() {
-	h, err := NewHub(viper.GetViper())
+// WithTransportURL sets the transport to use by parsing the provided URL.
+func WithTransportURL(tu string) Option {
```
sagikazarmark, line 105: TBH, I like this approach even less. IMHO the Hub itself should not be concerned about how a transport is constructed. It should rely solely on the `Transport` interface. Forcing to use a URL makes the library less reusable in my opinion.

hub/hub.go
```diff
@@ -4,80 +4,186 @@ import (
...
+		log.Panic(err)
+	}
+	h.transport = t
+
+	return h
+}
+
+// Stop stops disconnect all connected clients.
+func (h *Hub) Stop() error {
+	return h.transport.Close()
```
sagikazarmark, line 188: If we consider the transport an external dependency, the `Close` method should be called by the `main` function.

I would actually remove the `Close` method from the interface entirely, because it shouldn't be the Hub's concern.

Then in the main function:

```
transport, err := createTransportFromURL(u)
if err != nil {
    panic(err)
}

// This is only required so that createTransportFromURL can return an interface
// if in the main someone constructs a concrete transport type
// then the type assertion won't be necessary
if closer, ok := transport.(interface{ Close() error}); ok {
    defer closer.Close()
}

hub := hub.New(WithTransport(transport))
```

hub/metrics.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 // Metrics store Hub collected metrics.
+// TODO(dunglas): check how to let Caddy collecting these metrics: https://caddyserver.com/docs/metrics
```
sagikazarmark, line 11: IMHO metrics should be an interface. Then one can easily implement any kind of metrics collection (eg. implement an OpenTelemetry based metrics collection). That would increase the reusability of this library.

sagikazarmark (commented):
TBH it's not immediately obvious to me what `use the Caddy server module or the standalone library instead`. Should I register the HTTP handlers on the `Hub` struct in my own router? or is it going to implement the `http.Handler` interface?

hub/config.go
```diff
@@ -110,3 +124,115 @@ func InitConfig(v *viper.Viper) {
...
+	}
+	if p := v.GetStringSlice("publish_allowed_origins"); len(p) > 0 {
+		options = append(options, WithPublishOrigins(p))
+	}
+	if t := v.GetString("transport_url"); t != "" {
+		options = append(options, WithTransportURL(t))
+	}
+
+	h := New(options...)
+	h.config = v
```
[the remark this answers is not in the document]
dunglas: Yes, in the next version.

hub/hub.go
dunglas: I'll write the Caddy module and it will help to see the use case and if a refactoring is necessary.

sagikazarmark: Consider the following use case: I want to integrate mercure into an existing server process (because I don't need microservices, I'm perfectly fine with a single process). I already have a postgres connection open for my application, I want to reuse it for my mercure transport as well.

IMHO that's a better use case showcasing the reusability of mercure as a library than the Caddy module.

dunglas: @sagikazarmark this use case is already supported. Just register your own factory:

```go
func main() {
        pgCon := ...
	mercure.RegisterTransportFactory("pg", func newBoltTransport(u *url.URL, l Logger) (Transport, error) {
            // use pgCon here
        })
       h := mercure.NewHub(mercure.WithTransport(&url.URL{Scheme: "pg"}))
       defer h.Stop();
       // register the Mercure handler in your server
}
```

The current approach as another benefit. If you create a package calling `RegisterTransportFactory` in an `init()` function, the transport is automatically available in the Caddy module (by passing the appropriate `transport_url`in the `Caddyfile`): https://github.com/dunglas/mercure/pull/393/files#diff-be49a2dcca4004e34971b1e8dedc623412b90f41fbc7c18e1488cb73c73ec90fR96

As you probably know, I sell extra transports as part of the "on promise" version, so this feature is very important for me, but actually this will be helpful to everybody: create your transport, run:

```
xcaddy build \
      --with github.com/user/your-transport \             
      --with github.com/dunglas/mercure
```

And you have a Caddy build supporting Mercure and using your own transport!

This can definitely be improved (for instance the transport registry could be in another module, and/or we can expose a `WithTransport` function to bypass the factory if necessary), but IMHO it can be done in a next step.

WDYT?

sagikazarmark: > this use case is already supported. Just register your own factory:

I think registering a global transport factory with a local DB connection is a bad idea and is unnecessarily complicated.

> the transport is automatically available in the Caddy module

I don't see why the two can't co-exist though. Having a global factory registry doesn't mean that you have to call the factory inside `Hub`. IMHO the caddy module should call it and register the `Close` method in a shutdown hook instead of calling it in `Hub.Stop`.

> This can definitely be improved (for instance the transport registry could be in another module, and/or we can expose a WithTransport function to bypass the factory if necessary), but IMHO it can be done in a next step.

Yeah, it can definitely be part of a separate PR, this is quite huge already. 🙂

hub/hub.go
dunglas: fixed using a new struct, thanks for the tip.

hub/metrics.go
dunglas: Yes why not 👍, PR welcome on this too!

hub/hub.go
dunglas: Refactored! Thanks for the feedback.

sagikazarmark: Looks great, thanks!

hub/metrics.go
sagikazarmark: Will do, once this PR is merged.

metrics.go
```diff
@@ -39,26 +46,43 @@ func NewMetrics() *Metrics {
 			[]string{"topic"},
 		),
 	}
+
+	if err := m.registry.Register(m.subscribers); err != nil {
```
dunglas, line 50: This doesn't work. Even if `unregister()` has been called, it's not possible to register the same metric twice...

dunglas: And because how Caddy works, we need to find a workaround because when the config change (and in many other cases), this method is called again.

sagikazarmark: IMHO `Metrics` should be a dependency of the Hub as well. Then you can create it upfront.

handler.go
```diff
@@ -265,7 +265,7 @@ func (h *Hub) metricsHandler() http.Handler {
 	router := mux.NewRouter()
 
 	registerHealthz(router)
-	h.metrics.Register(router.PathPrefix("/").Subrouter())
+	h.metrics.(*PrometheusMetrics).Register(router.PathPrefix("/").Subrouter())
```
sagikazarmark, line 268: How about this?

```
if m, ok := h.metrics.(interface{ Register(r *mux.Router) }); ok {
    m.Register(router.PathPrefix("/").Subrouter())
}
```

metrics.go
```diff
@@ -10,38 +10,47 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-// Metrics store Hub collected metrics.
-type Metrics struct {
+type Metrics interface {
+	// ConnectSubscriber collects metrics about about subscriber connections.
+	ConnectSubscriber(s *Subscriber)
```
sagikazarmark, line 15: From a semantic point of view, I think these are events you want to subscribe to, so `SubscriberConnected`, `SubscriberDisconnected` and `UpdatePublished` sound better to me.

publish.go
```diff
@@ -69,5 +67,7 @@ func (h *Hub) PublishHandler(w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, u.ID)
 	h.logger.Info("Update published", zap.Object("update", u), zap.String("remote_addr", r.RemoteAddr))
 
-	h.metrics.NewUpdate(u)
+	if h.metrics != nil {
```
sagikazarmark, line 70: I think having a `noopMetrics` implementation and falling back to it in the constructor when no metrics is set using `WithMetrics` would be slightly better than checking `h.metrics` every time.

metrics.go
dunglas: I was hesitating between the two forms, so let's do this, thanks.

handler.go
dunglas: This function will be dropped in the next version, so I'm not sure it is worth it (it could encourage users to use it).

publish.go
dunglas: Sure, it was a remainder of the last iteration.

handler.go
sagikazarmark: The downside of the current solution is that it will panic if the injected metrics is not of type `*PrometheusMetrics` (which can very well be the case since the Hub accepts an interface now).

I know it will be dropped, so it doesn't matter that much, but it'll keep the code bug free if anyone tries to use this intermediate version (if it ever hits master in this form).

sagikazarmark: I missed that his code path is already behind a type assertion above. As long as that's the case, I guess this is fine.

sagikazarmark:
🎉

PR #399, opened 2020-11-25
fix: switch to form3tech's fork of jwt-go

### LINKED ISSUE #394: High severity issue in github.com/dgrijalva/jwt-go
There is an unpatched, high severity issue in the aforementioned JWT package: dgrijalva/jwt-go#428

Unfortunately, it looks like the author completely abandoned the package.

There is a maintained fork that fixes the issue, but a lot of users seem to prefer gopkg.in/square/go-jose.v2 now.

dunglas:
Closes #394.


sagikazarmark:
Looks like this fork has its own issues: https://github.com/go-kit/kit/issues/1026#issuecomment-720597430

form3tech-oss/jwt-go#5

The problem is that the token might contain an audience claim in which case the library could simply panic.

PR #415, opened 2020-12-09
Authorization: add code examples

alexislefebvre:
I took examples from Symfony's documentation: https://symfony.com/doc/4.4/mercure.html#authorization

### DIFF
--- a/docs/spec/faq.md
+++ b/docs/spec/faq.md
@@ -51,12 +51,12 @@ To unsubscribe, the client just calls `EventSource.close()`.
 
 Also, Mercure can easily be integrated with Apollo GraphQL by creating [a dedicated transport](https://github.com/apollographql/graphql-subscriptions).
 
-## How to send authorization to Mercure?
+## How to Send the Authorization Cookie to the Hub?
 
-Cookies are automatically sent by the browsers when opening an `EventSource` connection if the `withCredentials` attribute is set to `true`:
+Cookies are automatically sent by the browser when opening an `EventSource` connection if the `withCredentials` property is set to `true`:
 
 ```javascript
-const eventSource = new EventSource(hub, {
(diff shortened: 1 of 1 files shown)


dunglas:
Hi, thanks for your contribution. In my opinion this shouldn't be in the spec itself. The spec shouldn't reference non-standard features such as passing custom headers with `EventSource` and don't have to be exhaustive regarding other use cases covered by other specs such as EventSource (implementers must refer to the linked spec instead).

Would you mind to move these examples in [the FAQ](https://github.com/dunglas/mercure/tree/main/docs/spec/faq.md) instead?

alexislefebvre:
@dunglas hi, I moved the text to the FAQ.

docs/spec/faq.md
```diff
@@ -50,3 +50,24 @@ Updates for the given subscription can then be sent from the GraphQL server to t
...
+
+Cookies are automatically sent by the browsers when opening an `EventSource` connection if the `withCredentials` attribute is set to `true`:
+
+```javascript
+const eventSource = new EventSource(hub, {
+    withCredentials: true
+});
+```
+
+If the client is not a web browser, cookies can't be used.
```
dunglas, line 64: This should be removed, most popular HTTP clients including `curl` also support cookies.

docs/spec/faq.md
```diff
@@ -50,3 +50,24 @@ Updates for the given subscription can then be sent from the GraphQL server to t
...
+In order to achieve that, use a [polyfill](https://github.com/Yaffle/EventSource):
```
dunglas, line 65: Actually the polyfill is only useful for very old browsers and for cross-domain authorization.

dunglas (changes requested):
We are almost there, thanks!

dunglas (approved):

PR #422, opened 2020-12-16
docs: debug the Mercure.rocks hub

### LINKED ISSUE #388: Memory leaking
Hello! 

I am observing strange mercure behaviour. It has around 1000 or less subscribers but it seems that memory consumption increases gradually until it eats up all available memory. Then processor usage sky-rocket and situation does not improve until restarted.

Zabix screenshot

I started from 1 core and 2 gigs of ram. Then upped to 2 cores and 4 gigs. Now it runs on 2 cores and 8 gigs of ram. Behaviour seems repeat no matter the amount of ram we give it.

After restart situation gets back to normal.

Zabix screenshot

I'm running official docker installation dunglas/mercure:v0.10.4

What should be expected memory usage? What can be done to debug the situation?

dunglas:
Explain how to profile the hub. This should help for #388.

### DIFF
--- a/docs/hub/debug.md
+++ b/docs/hub/debug.md
@@ -1,0 +1,32 @@
+# Debug the Mercure.rocks Hub
+
+The hub is shipped with [`pprof`](https://blog.golang.org/pprof),
+a profiler allowing to find bottlenecks, memory leaks and blocked goroutines 
+among other things.
+
+To enable the profiler, add the `debug` global directive to your `Caddyfile`:
+
+```Caddyfile
+{
+    debug
+}
+
+# ...
+```
+
+The route exposing profiling data is now available at `http://localhost:2019/debug/pprof/`.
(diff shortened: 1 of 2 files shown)


docs/hub/debug.md
```diff
@@ -0,0 +1,32 @@
...
+
+The hub is shipped with [`pprof`](https://blog.golang.org/pprof),
+a profiler allowing to find bottlenecks, memory leaks and blocked goroutines 
+among other things.
+
+To enable the profiler, add the `debug` global directive to your `Caddyfile`:
+
+```Caddyfile
+{
+    debug
```
aurelijusrozenas, line 11: You must add `admin` not `debug`, that worked for me and caddy docs tells:
> debug enables debug mode, which sets all log levels to debug (unless otherwise specified).

And if you want it to work with docker you must enabled it like this `admin 0.0.0.0:2019`.

PR #426, opened 2020-12-18
feat: Support PKCS1/PKCS8 certificate for RSA

### LINKED ISSUE #419: Can't get RS256 to work
Hi, I wanted to try out the new release with the namespace fallback addition and can't get RS256 to work.
My suspicion was that I'm not passing in the multiline key properly, but I've tried multiple ways… 

My key and token is valid according to jwt.io. I also checked out the jwt-go repo and used my key and token in the tests and they pass OK.

I've read all the docs and can't get it right – This is what I think should work as caddy doc says everything between quotes is preserved:

```
        subscriber_jwt "
-----BEGIN CERTIFICATE-----
MIIDETCCAfmgAwI…
-----END CERTIFICATE-----
" RS256

```

And with legacy-v11 (docker-compose):
```
    environment:
            SUBSCRIBER_JWT_ALGORITHM: RS256
            SUBSCRIBER_JWT_KEY: |
                -----BEGIN CERTIFICATE-----
                MIIDETCCAfmgAwI…
                -----END CERTIFICATE-----
```

Both result in the same error:

```
unable to parse JWT: unable to parse PKIX public key: asn1: structure error: tags don't match (6 vs {class:2 tag:0 length:3 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} ObjectIdentifier @2
```

When I go back to v10, it complains about passing array in aud claim, maybe that happens _after_ public key is already parsed?

I'm using HS256 for publishing and that works without a problem.

Any help appreciated 🙇‍♂️

TamasSzigeti:
Sorry for not looking before, as the jwt-go package tests passed with my auth0 setup, I had the wrong assumptions and missed the fact that mercure is parsing public keys and not certificates.

I'm using the jwt-go parser which supports both so mercure can support auth0 that provides certificate and not a public key.

This also means the error messages have changed because of this fallback, I hope that's fine.

Fixes #419


dunglas:
Good catch, thank you very much for finding and squashing this limitation!

authorization_test.go
```diff
@@ -16,6 +16,7 @@ const (
...
+	validFullHeaderRsaForCert    = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkxIbzlPMmNNUzBqbzRsQWwtRk11ayJ9.eyJodHRwczovL21lcmN1cmUucm9ja3MvIjp7InB1Ymxpc2giOlsiZm9vIiwiYmFyIl0sInN1YnNjcmliZSI6WyJmb28iLCJiYXoiXX0sImlzcyI6Imh0dHBzOi8vbWVyY3VyZS10ZXN0LmV1LmF1dGgwLmNvbS8iLCJzdWIiOiJhdXRoMHw1ZmRjM2U4OGUzYjA0YjAwNzZhNTQxM2MiLCJhdWQiOlsiaHR0cHM6Ly9tZXJjdXJlLnJvY2tzL2FwaSIsImh0dHBzOi8vbWVyY3VyZS10ZXN0LmV1LmF1dGgwLmNvbS91c2VyaW5mbyJdLCJpYXQiOjE2MDgyNzAxMTQsImV4cCI6MTYwODM1NjUxNCwiYXpwIjoiMzAxaHptQkEydnlnN2dKVmJIRUxSVEN6WXR1QmtVTnYiLCJzY29wZSI6Im9wZW5pZCBwcm9maWxlIGVtYWlsIn0.WhMkaOvIckY7PFCYs5SvIcl8OK32z7AhCCPGx0G3yF4L0nOTssXV9gAPEpEOrLbCOG3ALDxOGB4VagGnwBIYuztBsuZyPRoIwUkBZOgrIUJYS96jYTb9osPUYZ7BxNlVefFse93JmeSFTZRi6oH5lbqCEW6FUVKNlHWBBl39UK9Fg36EFtOHIJ7wZ_NX51TPvN-roCp27qIhY3atDcHYWXTKS7VjznKDKxl5G7AmyA1L7eE3vpnGiECFcLwxr9BJGVQmnPnwUKf-tY3pSBY0gFE562y15vXk3D2il43uAP4uK2sw8rcup_CmWnT3wmAdIODkwdn8zKM3GC3Y-9WEvQ"
```
dunglas, line 19: This token contains an `exp` claim and has expired, this breaks the tests. Could you provide a token without this claim please @TamasSzigeti?

TamasSzigeti: I was going to but I see you already tackled it (: Thank you

dunglas: Fixed by #428, don't bother @TamasSzigeti!

PR #432, opened 2021-01-04
spec: fix the "dart_mercure" link

meyerbaptiste:


dunglas:
Good catch, thanks Baptiste!

PR #437, opened 2021-01-07
feat: Add /health endpoint

BafS:
Useful for probes


dunglas:
Why not! Can you rename it `/healthz`? We used `z` URLs in previous versions, let's be consistent to make upgrades easier.

BafS:
Great thanks, I updated the PR

dunglas:
Thanks!

PR #457, opened 2021-01-22
docs: fix: UI endpoint + add in into directives detail

silverbackdan:
fix: UI endpoint in docs
docs: add ui path into directives detail 

It took me a while to find this endpoint and checked the source to figure out once had found it in the docs it needs the final slash to load (at least in the current version). Thought it may be useful to re-iterate as well as fix the typo.


dunglas (approved):

dunglas:
Thanks!!

PR #471, opened 2021-02-21
Fix hanging goroutines when closing a transport

fabpot:
When closing a tranport via `Close()`, the `Subscriber::start()` and `Subscribe::SubscribeHandler()` goroutines hang. Closing `live.in` and `out` allows to end these goroutines properly.


dunglas (approved):
Good catch

dunglas:
The test failure looks related.

PR #492, opened 2021-04-01
ci: simplify golangci-lint config to prevent random failures

dunglas:


dunglas:
Closing in favor of #493

PR #503, opened 2021-04-26
fix!: don't partition the Prometheus metrics per topic to prevent a memory leak

### LINKED ISSUE #463: Subscription API causes memory overshoot
Hi. I'm using Mercure's official docker image with a few customizations in Caddyfile. Without the subscription API turned on, the memory usage is very low. But as soon as I enable the subscription API, the memory usage jumps to 100% and the container crashes. Container and load specs are as follows:

CPU units: ~680 (t3a.medium EC2 instance, 2 vCPUs, shared by one Mercure task with two other tasks)
Memory: 256
Subscribers: ~50
Topics: ~5
Messages: 5/minute (avg)

The memory graph attached shows the memory used when subscription API is enabled and when it's disabled. From 10:05 to 11:45 is the time when subscription API is disabled. From 07:00 to 10:00 and from 11:45 onwards are the windows when subscription API is enabled. The dips you see in those windows are actually crashes, not low memory usages. The memory consupmtion is constantly 100% when the container is running.

Mercure memory

Unfortunately there's nothing in cloudwatch logs before the crash happens, and the reason for task being stopped in ECS console (strangely enough) is not out of memory. It says that the essential container in task exited.

dunglas:
Closes #463.

As the number of topics is not known in advance, partitioning the Prometheus metrics introduces a memory leak: the number of gauge and counter can grow forever, and quickly.

According to https://github.com/prometheus/client_golang/issues/791#issuecomment-673538744, Prometheus cannot/shouldn't be used for this kind of metrics, and the user should rely on log analysis instead.

@divine could you try is this patch fixes your issue with the Caddy version please?
@jdecool @sagikazarmark as you contributed the Prometheus support, could you take a look if this patch is legit?

Thank you both!

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -25,9 +25,9 @@ func (NopMetrics) UpdatePublished(s *Update)            {}
 // PrometheusMetrics store Hub collected metrics.
 type PrometheusMetrics struct {
 	registry         prometheus.Registerer
-	subscribersTotal *prometheus.CounterVec
-	subscribers      *prometheus.GaugeVec
-	updatesTotal     *prometheus.CounterVec
+	subscribersTotal prometheus.Counter
+	subscribers      prometheus.Gauge
+	updatesTotal     prometheus.Counter
 }
 
 // NewPrometheusMetrics creates a Prometheus metrics collector.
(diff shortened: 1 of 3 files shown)


divine:
Hello,

I have deployed the Caddy version with this commit, I will report back if something breaks.

Thank you for the quick fix!

____
Not related to this PR:

Also, for this issue:

> The logger (Zap), and especially the calls to zap.Object() when logging a subscriber

https://github.com/dunglas/mercure/blob/16cc2402d5eab38c72a9493bc60d80c9c4d0a435/subscribe.go#L88
Can this moved behind some function and if the error level is equal to info then do some logging?

I'm not a Go developer but I think this might waste some resources, sorry if it otherwise (taking into account 2M notifications per day).

Thanks!

sagikazarmark:
Looks good to me.

jdecool:
LGTM to me too

dunglas:
Thanks!

dunglas:
@divine it's how Zap works under the hood. This will be logged only if the level is superior or equal to INFO.

PR #512, opened 2021-04-30
feat: add a Helm chart

dunglas:
Add a new Helm chart for the Caddy build. Fixes https://github.com/dunglas/mercure/issues/357.

@sagikazarmark, I hesitated to use your Caddy chart, but as the hub has some specificities, I don't think that it's worth it, so I created a more specific one. However, I'll be glad to collaborate with you to share everything that can be shared.


sagikazarmark:
@dunglas sure, a specific chart is better in my opinion.

I think using [helm-docs](https://github.com/norwoodj/helm-docs) is a good way to generate documentation from the values file, so I'd consider adding that.

I'd also consider grouping application specific configuration under a `config` key, just to make it more separated from the rest of the chart.

Updating the ingress template is also worth doing: https://github.com/sagikazarmark/helm-charts/blob/master/charts/caddy/templates/ingress.yaml
I already have PR that will hopefully will be merged in Helm 3.6: helm/helm#9621

helm/mercure/templates/deployment.yaml
```diff
@@ -0,0 +1,103 @@
...
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: http
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: http
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
```
nitneuk, line 91: By default, k8s sends a SIGKILL to stop the containers : https://kubernetes.io/fr/docs/concepts/workloads/pods/pod/
From what I see here https://caddy.community/t/does-taskkill-allow-caddy-to-shut-down-gracefully/1099 and here  https://caddyserver.com/docs/api#post-stop 
Maybe call the API to make a graceful shutdown, or using a SIGINT would be better : 
```suggestion
            {{- toYaml .Values.resources | nindent 12 }}
          lifecycle:
            preStop:
              exec:
                command: ["curl", "-XPOST", "http://localhost:2019/stop"]
```

nitneuk (approved):
LGTM

helm/mercure/templates/deployment.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "mercure.fullname" . }}
```
nitneuk, line 4: Personally, I suffix the resource name with the kind in lowercase as made here : https://helm.sh/docs/chart_template_guide/named_templates/#declaring-and-using-templates-with-define-and-template

vincentchalamon: As it is the main deployment, the name would be weird or hard to define: `mercure-mercure`, `mercure-deployment`, etc.
In this special case, I don't recommend to specify a suffix

helm/mercure/templates/ingress.yaml
```diff
@@ -0,0 +1,41 @@
...
+    {{- end }}
+  {{- end }}
+  rules:
+    {{- range .Values.ingress.hosts }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+          {{- range .paths }}
+          - path: {{ .path }}
+            backend:
```
vincentchalamon, line 36: On k8s cluster >=v1.19.0, the ingress chart has changed:
```yaml
- path: {{ . }}
  {{ if semverCompare ">=1.19-0" $kubeVersion }}
  pathType: Prefix
  {{ end }}
  backend:
    {{ if semverCompare ">=1.19-0" $kubeVersion }}
    service:
      name: {{ $fullName }}
      port:
        number: {{ $svcPort }}
    {{ else }}
    serviceName: {{ $fullName }}
    servicePort: {{ $svcPort }}
    {{ end }}
```

helm/mercure/templates/ingress.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if .Values.ingress.enabled -}}
+{{- $fullName := include "mercure.fullname" . -}}
+{{- $svcPort := .Values.service.port -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
```
vincentchalamon, line 4: v1.14 is not supported anymore (cf. [release notes](https://cloud.google.com/kubernetes-engine/docs/release-notes)):
```yaml
{{- if semverCompare ">=1.19-0" $kubeVersion -}}
apiVersion: networking.k8s.io/v1
{{- else -}}
apiVersion: networking.k8s.io/v1beta1
{{- end }}
```

vincentchalamon (changes requested):

helm/mercure/templates/deployment.yaml
sagikazarmark: I don't recommend that either.

helm/mercure/templates/ingress.yaml
sagikazarmark: I suggest following the official template: helm/helm#9621

helm/mercure/templates/deployment.yaml
```diff
@@ -0,0 +1,103 @@
...
+  labels:
+    {{- include "mercure.labels" . | nindent 4 }}
```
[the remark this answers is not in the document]
sagikazarmark: I suggest following the official Helm template unless there is a reason not to. It helps updating manifests in the long run.

helm/mercure/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{/*
+Common labels
+*/}}
+{{- define "mercure.labels" -}}
+helm.sh/chart: {{ include "mercure.chart" . }}
+{{ include "mercure.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
```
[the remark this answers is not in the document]
dunglas: Indeed, I think that the missing ones don't apply here.

charts/mercure/templates/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+                  name: {{ include "mercure.fullname" . }}
+                  key: subscriber-jwt-alg
+            - name: MERCURE_EXTRA_DIRECTIVES
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "mercure.fullname" . }}
+                  key: extra-directives
+          {{- if .Values.persistence.enabled }}
+          volumeMounts:
+            - mountPath: /data
```
sagikazarmark, line 80: I'm more conservative on these issues, so I'd vote for `/var/mercure`.

charts/mercure/templates/ingress.yaml
```diff
@@ -0,0 +1,61 @@
...
+                name: {{ $fullName }}
+                port:
+                  number: {{ $svcPort }}
+              {{- else }}
+              serviceName: {{ $fullName }}
+              servicePort: {{ $svcPort }}
+              {{- end }}
+          {{- end }}
+    {{- end }}
+  {{- end }}
```
sagikazarmark, line 61:
```suggestion
{{- end }}
```

This needs to be fixed in my caddy chart as well.

charts/mercure/values.yaml
```diff
@@ -0,0 +1,129 @@
...
+# -- the JWT algorithm to use for publishers
+publisherJwtAlg: HS256
+
+# -- the JWT key to use for subscribers, a random key will be generated if empty
+subscriberJwtKey: ""
+# -- JWT algorithm to use for subscribers
+subscriberJwtAlg: HS256
+
+# -- license key for [the High Availability version](https://mercure.rocks/docs/hub/cluster) (not necessary is you use the FOSS version):
+license: ""
```
sagikazarmark, line 25: I still think these should be under a `config` key to better separate them from the rest of the chart values.

dunglas: Is it a common practice? AFAIK most charts don't do that (and it looks in contractions with Helm's best practices: https://helm.sh/docs/chart_best_practices/).

sagikazarmark: I don't see a contradiction with the best practices. I don't think the claim about the mandatory check in case of nesting is true.

I honestly can't tell whether most charts do it or not. It certainly proved useful for me. Take it with a grain of salt.

charts/mercure/values.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - host: mercure-example.local
+      paths:
+      - path: /
+        backend:
+          serviceName: mercure-example.local
+          servicePort: 80
+  tls: []
+  #  - secretName: mercure-example-tls
+  #    hosts:
+  #      - mercure-example.local
```
sagikazarmark, line 81: This needs to be updated as well.

dunglas: Good catch, thanks! I also shamelessly copied your docs for the values that weren't documented.

sagikazarmark: No problem! 😄

Issue #535, opened 2021-07-15
Missing Last-Event-Id from CORS allowedHeaders


robertodormepoco:
Hi, I'm currently having problems using a retry aware js SSE client, when it tries to catch up after a disconnection it gets a HTTP 403 during preflight due to the presence of the Last-Event-Id header in the Access-Control-Request-Headers directive (valid as of the RFC and Mercure docs)
i.e.
```
Access-Control-Request-Method: GET
Access-Control-Request-Headers: authorization,cache-control,last-event-id
```

removing last-event-id does the job, though i was not even confident the problem was in the caddy mercure module, more on the Caddyfile (which is the default one from the official Dockerfile), so i looked somewhere else with no luck, in the end i just had to check the code to be sure and I've found this

https://github.com/dunglas/mercure/blob/6788c41ab1395bc465fe7bf2de20ec7ad9cb8412/handler.go#L69

is this correct? I'm missing something here?

Thanks

robertodormepoco:
fixed in #538

PR #538, opened 2021-07-21
fix: Add Last-Event-Id as CORS allowed header

robertodormepoco:
fix: Add Last-Event-Id as CORS allowed header

see #535


dunglas (approved):
LGTM, thanks!

robertodormepoco:
@dunglas bump, just to let the workflows go forward :)

dunglas:
Thanks @robertodormepoco!

Issue #532, opened 2021-07-05
Feature Request: config option to customize cookie name


dominicbvm:
In our dev environment, we customize our hosts file so we can use the same parent domain and wildcard SSL cert that we use in production. This causes problems if we are jumping between dev and prod because they both set a wildcard cookie for the same parent domain to auth to mercure but different mercure instances. It would be nice to be able to customize the cookie name in a scenario like this so that it could be something other than mercureAuthorization.

dunglas:
This will also allow to create different security contexts on the same domain.

I plan to add this feature to the spec and to implement it in the hub.

dunglas:
I opened a proposal to add support for this feature to the spec. Could you review it to check if it fixes this issue please? https://github.com/dunglas/mercure/pull/562

dominicbvm:
Yeah, reading through the spec changes around the cookie name, this should definitely solve the scenario we are running into. Thanks!

Issue #387, opened 2020-11-02
Authentication with multiple JWT tokens from the same browser session


damijank:
I have a question about serving multi-tenant / multi-client apps.

We have a setup like this one:
* a single multi-tenant API Platform backend (on `paas.org`)
* a single Mercure hub, which is on the same domain as PaaS
* a multi-client-enabled CMS (on `tenant.org` domain)
  * the CMS is allowed to administrate all the clients for their tenant 
* multiple clients (websites) are on their own domains, too (`client-one.org`, `client-two.org`...)

_For reference:
in this setup the **tenants** (our customers) are parents of **clients** (our customers have multiple websites, mobile apps, etc.)._

When working just with a single client, we ran into the problem that the client and the hub are not on the same second-level domain. I've seen #17 and the polyfill solution is not what we would want since that falls back to plain ol' HTTP and as such is not as "native" and performant as the real native `EventSource`.
We could/should fix this with the PaaS setting the `mercureAuthentication` cookie at the time of response to OAuth, and since the hub is on the same domain this should work... almost, I'm not going into detail with OAuth scopes. But still, even that is not the whole story. :slightly_smiling_face: 

When switching to a second client in the CMS, it has to fetch a different JWT from PaaS (which would reset the cookie), then the CMS opens a new `EventSource` connection to the hub with the new JWT.
With the solution when the PaaS sets the mercure cookie, we here _loose_ the `EventSource` connection for the first client.
If we exposed the header to JS, it still wouldn't work because JS can not set a cookie for another domain (it's not a subdomain, so a `Domain` attribute isn't suitable either).

We want to enable the customer to have multiple tabs with CMS open in the same browser session, each administering it's own client. With cookie-based mercure authentication we can not have each tab receiving it's own SSEs because the cookie can only be set to a single value.

I've found this #150 and I understand the goal of securing the JWT. But in the end, with HTTPS the URI itself is encrypted in transport. I'd suggest to add a query parameter as a possibility for mercure authentication. The current limitation is restrictive enough so we're not able to accomplish what we've aimed for.

I'm open to any other suggestions which would resolve this problem, without the need to transport the JWT in a query parameter. Please advise!

damijank:
For now, I've made a workaround via a reverse-proxy configuration in my nginx
```
location /.well-known/mercure {

    ...

    # TODO: Reverse-proxy workaround for: https://github.com/dunglas/mercure/issues/387
    set $fw_cookie "";
[...]

PR #562, opened 2021-09-17
spec: allow configuring cookie name and using authorization query parameter

dunglas:
See https://github.com/dunglas/mercure/issues/387 and https://github.com/dunglas/mercure/issues/532.

Some sentences have been adapted from the OAuth RFC: https://datatracker.ietf.org/doc/html/rfc6750

### DIFF
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -97,7 +97,7 @@ The publisher **MAY** provide the following target attributes in the Link Header
     will be pushed, using formats such as JSON Patch [@RFC6902] or JSON Merge Patch [@RFC7386].
 *   `cookie-name`: the name of the cookie that contains the JWS [@!RFC7515] used for authorization
     ((#authorization)). If this attribute is not set, the name of the cookie is assumed to be
-    `mercureAuthorization`. Using this attribute allows prevents cookie name conflicts when using
+    `mercureAuthorization`. Using this attribute allows to prevent cookie name conflicts when using
     multiple hubs on the same domain.
 *   `key-set`: the URL of the key set to use to decrypt updates, encoded in the JWK set format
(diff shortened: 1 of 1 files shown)


spec/mercure.md
```diff
@@ -88,18 +81,20 @@ header). The target URL of these links **MUST** be a hub implementing the Mercur
...
- *  `content-type`: the content type of the updates that will be pushed by the hub. If omitted,
+*   `content-type`: the content type of the updates that will be pushed by the hub. If omitted,
     the subscriber **MUST** assume that the content type will be the same as that of the original
     resource. Setting the `content-type` attribute is especially useful to hint that partial updates
     will be pushed, using formats such as JSON Patch [@RFC6902] or JSON Merge Patch [@RFC7386].
-
- *  `key-set`: the URL of the key set to use to decrypt updates, encoded in the JWK set format
+*   `cookie-name`: the name of the cookie that contains the JWS [@!RFC7515] used for authorization
+    ((#authorization)). If this attribute is not set, the name of the cookie is assumed to be
+    `mercureAuthorization`. Using this attribute allows prevents cookie name conflicts when using
```
damijank, line 95: > Using this attribute allows **to prevent** cookie name conflicts...

dunglas: Good catch. It should even be `preventing`.

damijank: Yeah, was zig-zagging between the two, too... number of Google resuts for each phrase ([12.5M](https://www.google.com/search?q=%22allows+to+prevent%22) vs. [16.9k](https://www.google.com/search?q=%22allows+preventing%22)) and [this comparison](https://textranch.com/387138/allows-to-prevent/or/allows-preventing/) made me choose the "to prevent" version. :man_shrugging:

PR #578, opened 2021-11-01
perf: Subscriber performance

kevburnsjr:
Significantly improves memory usage and throughput with large number of subscribers.


dunglas:
Thank you very, very much @kevburnsjr!

dunglas:
Humm, it looks like we may have a data race @kevburnsjr: https://github.com/dunglas/mercure/runs/4192750294?check_suite_focus=true

subscriber.go
```diff
@@ -168,43 +98,91 @@ func (s *Subscriber) Receive() <-chan *Update {
...
 func (s *Subscriber) HistoryDispatched(responseLastEventID string) {
 	s.responseLastEventID <- responseLastEventID
-	close(s.history.in)
 }
 
 // Disconnect disconnects the subscriber.
 func (s *Subscriber) Disconnect() {
 	s.disconnectedOnce.Do(func() {
 		close(s.disconnected)
+		close(s.out)
```
dunglas, line 107: This change creates a deadlock. https://github.com/dunglas/mercure/pull/578/files#diff-802f9d2c1e1a0efff7a8474036a024fbbba988f5bdf4650a0b3b4dad99df9ca2R74 can be called after this, and block forever.

PR #587, opened 2021-11-15
fix: redact sensitive informations from logs

dunglas:
It's better to not log JWT.


darkweak (approved):

dunglas:
Replaced by https://github.com/caddyserver/caddy/pull/4438.

PR #615, opened 2022-02-21
spec: add Freddie in the implementation list

dunglas:
Adds @bpolaszek's [Freddie] in the implementations.


bpolaszek:
> Adds @bpolaszek's [Freddie] in the implementations.

Much appreciated. Thanks! 🙏

PR #619, opened 2022-02-22
spec: improve abstract

dunglas:

### DIFF
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -27,9 +27,14 @@ organization = "Les-Tilleuls.coop"
 
 .# Abstract
 
-Mercure is a protocol enabling the pushing of data updates to web browsers and other HTTP clients in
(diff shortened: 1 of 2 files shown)


spec/mercure.md
```diff
@@ -27,9 +27,14 @@ organization = "Les-Tilleuls.coop"
 
 .# Abstract
 
-Mercure is a protocol enabling the pushing of data updates to web browsers and other HTTP clients in
+Mercure provides a common publish-subscribe mechanism for public and private web resources.
+Mercure enables the pushing of any web content to web browsers and other clients in
```
alanpoulain, line 31:
```suggestion
Mercure enables the pushing of any content to web browsers and other clients in
```

dunglas: I reused the same working as in https://www.w3.org/TR/websub/.

PR #620, opened 2022-02-22
spec!: force the publisher JWT to contain the allowed topics

### LINKED ISSUE #579: Mercure publishes to topic JWT token should not be allowed to
### Mercure version: `0.13`

Without the "private" flag on messages are being sent to subscriber even if topic is not in subscriber or publish allowed whitelist.

image

Token:
`eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJtZXJjdXJlIjp7InB1Ymxpc2giOlsiL3dvcmtzcGFjZXMvMTIzIl0sInN1YnNjcmliZSI6WyIvd29ya3NwYWNlcy8xMjMiXX19.bxyELLWxI60AUT1xC3OR0cU_62vRUV0WVyuj00EDWjs`

Unauthorized is only thrown if "Private" is ticked. 
This does not work well for client-to-client mercure applications as this can be overwritten by any client.

dunglas:
Closes #579.

This is a BC break, but it improves the security of the protocol.
To allow a publisher to publish to all topics, a star (`*`) can be used.

### DIFF
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -417,12 +417,17 @@ To be allowed to publish an update, the JWS presented by the publisher **MUST** 
 called `mercure`, and this claim **MUST** contain a `publish` key. `mercure.publish` contains an
 array of topic selectors. See (#topic-selectors).
 
+<<<<<<< HEAD
 If `mercure.publish`:
 
 *   is not defined, then the publisher **MUST NOT** be authorized to dispatch any update
 *   contains an empty array, the publisher **MUST NOT** be authorized to publish private updates,
     but can publish public updates for all topics.
 
+=======
+If `mercure.publish` is not defined, then the publisher **MUST NOT** be authorized
+to dispatch any update.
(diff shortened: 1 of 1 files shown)


spec/mercure.md
```diff
@@ -386,13 +386,8 @@ To be allowed to publish an update, the JWS presented by the publisher **MUST**
...
 array of topic selectors. See (#topic-selectors).
 
-If `mercure.publish`:
-
- *  is not defined, then the publisher **MUST NOT** be authorized to dispatch any update
-
- *  contains an empty array, the publisher **MUST NOT** be authorized to publish private updates,
-    but can publish public updates for all topics.
-
+If `mercure.publish` is not defined, then the publisher **MUST NOT** be authorized
```
alanpoulain, line 389: or empty?

dunglas: I don't think that it's necessary to make it explicit, as it's not the case for `subscribe` too.

PR #652, opened 2022-06-16
Fix description of Transport.RemoveSubscriber

damz:
There is a small copy/paste issue in the description of `Transport.RemoveSubscriber`.


dunglas (approved):

dunglas:
Good catch thanks!

PR #653, opened 2022-06-16
spec!: rename the Last-Event-ID query parameter in lastEventID

### LINKED ISSUE #613: Use lastEvendID instead of Last-Event-ID as query parameter
To be in sync with the default parameter supported by the most popular EventSource polyfill: https://github.com/Yaffle/EventSource#custom-query-parameter-name-for-the-last-event-id

azjezz:
closes #613
closes #621

### DIFF
--- a/subscribe.go
+++ b/subscribe.go
@@ -76,7 +76,7 @@ func (h *Hub) SubscribeHandler(w http.ResponseWriter, r *http.Request) {
 
 // registerSubscriber initializes the connection.
 func (h *Hub) registerSubscriber(w http.ResponseWriter, r *http.Request) *Subscriber {
-	s := NewSubscriber(retrieveLastEventID(r), h.logger)
+	s := NewSubscriber(retrieveLastEventID(r, h.logger), h.logger)
 	s.Debug = h.debug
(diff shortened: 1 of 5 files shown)


subscribe.go
```diff
@@ -153,12 +153,18 @@ func sendHeaders(w http.ResponseWriter, s *Subscriber) {
...
 
 // retrieveLastEventID extracts the Last-Event-ID from the corresponding HTTP header with a fallback on the query parameter.
-func retrieveLastEventID(r *http.Request) string {
+func retrieveLastEventID(r *http.Request, logger Logger) string {
 	if id := r.Header.Get("Last-Event-ID"); id != "" {
 		return id
 	}
 
-	return r.URL.Query().Get("Last-Event-ID")
+	if id := r.URL.Query().Get("Last-Event-ID"); id != "" {
```
dunglas, line 161: nit: IMHO we should check if the key exists, even if it contains an empty string, to trigger the deprecation (JS code may always include this query parameter even if empty).

dunglas (approved):

azjezz:
tests are passing 👍

PR #655, opened 2022-06-17
feat: allow using authorization query parameter

### LINKED ISSUE #387: Authentication with multiple JWT tokens from the same browser session

azjezz:
closes #387

### DIFF
--- a/authorization.go
+++ b/authorization.go
@@ -34,6 +34,8 @@ const (
 var (
 	// ErrInvalidAuthorizationHeader is returned when the Authorization header is invalid.
 	ErrInvalidAuthorizationHeader = errors.New(`invalid "Authorization" HTTP header`)
+	// ErrInvalidAuthorizationQuery is returned when the authorization query parameter is invalid.
+	ErrInvalidAuthorizationQuery = errors.New(`invalid "authorization" Query parameter`)
 	// ErrNoOrigin is returned when the cookie authorization mechanism is used and no Origin nor Referer headers are presents.
 	ErrNoOrigin = errors.New(`an "Origin" or a "Referer" HTTP header must be present to use the cookie-based authorization mechanism`)
 	// ErrOriginNotAllowed is returned when the Origin is not allowed to post updates.
@@ -56,6 +58,14 @@ func authorize(r *http.Request, jwtConfig *jwtConfig, publishOrigins []string, c
 		}
 
 		return validateJWT(authorizationHeaders[0][7:], jwtConfig)
+	}
+
+	if authorizationQuery, queryExists := r.URL.Query()["authorization"]; queryExists {
+		if len(authorizationQuery) != 1 || len(authorizationQuery[0]) < 41 {
+			return nil, ErrInvalidAuthorizationQuery
+		}
+
+		return validateJWT(authorizationQuery[0], jwtConfig)
 	}
 
 	cookie, err := r.Cookie(cookieName)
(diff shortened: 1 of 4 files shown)


authorization.go
```diff
@@ -58,6 +60,15 @@ func authorize(r *http.Request, jwtConfig *jwtConfig, publishOrigins []string, c
 		return validateJWT(authorizationHeaders[0][7:], jwtConfig)
 	}
 
+	authorizationQuery, queryExists := r.URL.Query()["authorization"]
```
dunglas, line 63: You could probably merge with this line with the next one.

dunglas (approved):

authorization.go
azjezz: done.

dunglas:
Could you also add some config in `Caddyfile` to strip the value of the `authorization` query parameter from the logs? https://dunglas.fr/2022/04/caddy-logging-security-improvements/

authorization_test.go
```diff
@@ -207,6 +207,133 @@ func TestAuthorizeAuthorizationHeaderWrongAlgorithm(t *testing.T) {
...
+	claims, err := authorize(r, &jwtConfig{[]byte{}, jwt.SigningMethodHS256}, []string{}, defaultCookieName)
+	assert.EqualError(t, err, "unable to parse JWT: signature is invalid")
+	assert.Nil(t, claims)
+}
+
+func TestAuthorizeAuthorizationQueryInvalidKeyRsa(t *testing.T) {
+	r, _ := http.NewRequest("GET", defaultHubURL, nil)
+	query := r.URL.Query()
+	query.Set("authorization", validEmptyHeaderRsa)
+	r.URL.RawQuery = query.Encode()
```
azjezz, line 247: having to do this over and over is kinda of annoying :)

azjezz:
tests are passing 👍

dunglas (approved):

dunglas:
Thanks @azjezz

PR #657, opened 2022-06-20
feat(hub): rename the Last-Event-ID query parameter in lastEventID

azjezz:
closes https://github.com/dunglas/mercure/issues/613
ref https://github.com/dunglas/mercure/pull/621


dunglas:
The prefix of the commit message should be `feat(hub):`. Then we can merge!

dunglas:
Thanks Saif!

PR #660, opened 2022-07-05
feat: Add CADDY_EXTRA_DIRECTIVES

blackheaven:


dunglas (changes requested):
Good idea. Maybe just `EXTRA_DIRECTIVES` for consistency with `GLOBAL_OPTIONS` which doesn't include the CADDY prefix?

charts/mercure/templates/deployment.yaml
```diff
@@ -47,6 +47,11 @@ spec:
                 configMapKeyRef:
                   name: {{ include "mercure.fullname" . }}
                   key: global-options
+            - name: EXTRA_DIRECTIVES
+              valueFrom:
+                configMapKeyRef:
+                  name: {{ include "mercure.fullname" . }}
+                  key: extra-directives
```
dunglas, line 54: The config template must be updated as well isn't it? We also need a new value.

Also, I would use a secret as this env var may contains secret keys and things like this, depending on the modules you use.

WDYT?

blackheaven: I did it, but it may be confusing with `extra-directives`/`extraDirectives`, should I change it too?

dunglas (approved):

blackheaven:
Actually, it works on my computer, can you have a look please?

dunglas:
I'm not sure of waht the issue is, but the error in the CI is "wget: bad address 'mercure-jg4qdqb8ei:80'".

dunglas:
Thanks!

PR #670, opened 2022-07-21
feat(hub): add an option to support legacy versions of the protocol

dunglas:
Allow to enable or disable compatibility with old versions of the protocol (cc @azjezz).

### DIFF
--- a/hub.go
+++ b/hub.go
@@ -13,7 +13,7 @@ import (
 	"go.uber.org/zap"
 )
 
-// ErrUnsupportedProtocolVersion is returned when the version passed is unsupported
+// ErrUnsupportedProtocolVersion is returned when the version passed is unsupported.
 var ErrUnsupportedProtocolVersion = errors.New("compatibility mode only supports protocol version 7")
 
 // Option instances allow to configure the library.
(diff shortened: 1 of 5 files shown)


subscribe.go
```diff
@@ -163,6 +163,10 @@ func retrieveLastEventID(r *http.Request, logger Logger) string {
 		return id
 	}
 
+	if opt.protocolVersionCompatibility == 0 || opt.protocolVersionCompatibility >= 8 {
```
azjezz, line 166:
```suggestion
	if opt.protocolVersionCompatibility != 0 || opt.protocolVersionCompatibility < 8 {
```
we could do the opposite here and move the next if statement inside ( not to duplicate `return ""` )

hub.go
```diff
@@ -202,24 +219,25 @@ type jwtConfig struct {
...
+	dispatchTimeout              time.Duration
+	heartbeat                    time.Duration
+	publisherJWT                 *jwtConfig
+	subscriberJWT                *jwtConfig
+	metrics                      Metrics
+	allowedHosts                 []string
+	publishOrigins               []string
+	corsOrigins                  []string
+	cookieName                   string
+	protocolVersionCompatibility int
```
azjezz, line 240: I think it would a good idea to introduce `opt.IsProtocolVersionCompatibleWith(x)` to avoid repeating `if opt.protocolVersionCompatibility == 0 || opt.protocolVersionCompatibility >= 8` in the future ( it would be `opt.IsProtocolVersionCompatibleWith(7)` ).

dunglas: Done

subscribe.go
dunglas: I improved the logic to provide better debug information.

PR #741, opened 2023-01-15
Fix typo in sample command in install.md

JamuelStarkey:
fix: trivial typo in sample command to start mercure.


dunglas:
Oops! Thanks.

PR #748, opened 2023-02-07
feat: Add subscriberJwtSecret and publisherJwtSecret

gfr:
Add subscriberJwtSecret and publisherJwtSecret to reference existing secrets.


gfr:
We use a combination of vault and external-secrets.io for our secrets and we don't want to specify the secrets via values. Due to this combination our secrets already exists in the namespace where mercure is deployed and we need an option  to specify where our existing secret is located.

Issue #758, opened 2023-04-06
Timeout


ihab95:
Hello MercureTeam and the community,

We installed and integrated it on a Docker container, and now we are on the testing stage, near to go live.

All was OK until we've faced an incomprehensible bug and we need your help to understand : 

Mercure completely stopped to treat our data, impossible to push any message on Mercure hub. Our publisher workers are meeting a timeout. Mercure container is still up, and Caddy healthz route is responding a 200.

The only log we got is : 
`{"level":"debug","ts":1680709312.2403877,"logger":"http.handlers.mercure","msg":"write timeout: close the connection","subscriber":{"id":"urn:uuid:5fd605c2-7762-4819-839d-a3eca847488a","last_event_id":""...`

No way to know why our Mercure Hub is not callable.

dunglas:
That's weird. That may (?) be a deadlock? What happens if you restart the hub?

If you can reproduce, could you enable the debug mode to see what's going on? https://mercure.rocks/docs/hub/debug

ihab95:
We already restarted the hub, and it has solved the problem, but it will probably returns...
We're not able to reproduce actually, we are waiting for another occurence of the bug.

Thanks for the tip, we just enabled the debug mode, it could help us when it'll happen

mxmp210:
I can confirm something similar as well, if hub is running for long time(>1 Week ), it slows down significantly and requests just take forever to process - restarting hub resets everything. This has been happening quite few times recently on chat app I'm working on. We have quite bit of  updates that's happening every time.

The local transport is used so it is just a relay machine - no disk is involved in config (i.e. no bolt transport). In our case publishing and subscription happens but request / update times are > 6s to > 20s which suggests either caddy or mercure is having internal resource issues or we are missing something in config that's keeping hub busy ( which is I'm trying to investigate )

I was reluctant to report issue till it's figure out if it was config but it seems other users are having same outcome.
It seems there's internal memory leak that is it slowing down - rebooting mercure will reconnect all users(From client side re-connection methods) and you can see how less memory they use with same amount of load and number of connection connections.  Also this pattern is unpredictable because it happens after either certain number of connections or number of updates as hub get used - not sure which one yet.

![image](https://user-images.githubusercontent.com/24737091/230779780-cf9ce87d-47eb-4ca9-8c31-adae3754f89c.png)

dunglas:
[...]

PR #759, opened 2023-04-19
fix: don't block when subscriber chan is full

dunglas:
Should fix the bug found by @mholt in https://github.com/dunglas/mercure/issues/758#issuecomment-1513395883.

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -82,8 +82,9 @@ func (s *Subscriber) Dispatch(u *Update, fromHistory bool) bool {
 	case s.out <- u:
 		s.outMutex.Unlock()
 	default:
+		// Channel is full, disconnect the subscriber
+		atomic.StoreInt32(&s.disconnected, 1)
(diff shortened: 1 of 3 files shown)


subscriber.go
```diff
@@ -44,7 +45,7 @@ func NewSubscriber(lastEventID string, logger Logger) *Subscriber {
 		EscapedID:           url.QueryEscape(id),
 		RequestLastEventID:  lastEventID,
 		responseLastEventID: make(chan string, 1),
-		out:                 make(chan *Update, 1000),
+		out:                 make(chan *Update, 1000), // update TestSubscriberDoesNotBlockWhenChanIsFull if you change the buffer size
```
mholt, line 48: Maybe this size could be a `const` shared by the tests?

Anyway, LGTM! :) Thanks!

dunglas: Good idea, thank you

PR #772, opened 2023-05-24
fix(caddy): set default write timeout in transport URL

dunglas:


dunglas:
This can actually be handled transport-side, let's not increase the complexity for this.

Issue #546, opened 2021-08-04
Use of x-forwarder-for and metadata in "New subscriber" logs


si14:
First of all, thanks for this awesome piece of software!

We've deployed the latest Mercure and it works great. However, there are two tiny ops snags in its logging as you can see on the screenshot:

<img width="508" alt="Screenshot 2021-08-04 at 20 14 57" src="https://user-images.githubusercontent.com/316910/128241367-abc16c5d-efb0-4398-ab7a-04c51005267f.png">

Firstly, it uses IP of our LB, not the one from `X-Forwarder-For`. There is a very old issue about this that eventually got fixed https://github.com/dunglas/mercure/issues/114 , but I guess the fix was lost during the big rewrite.

Secondly, do you think it's possible to log JWT `payload` if it's present? According to the spec, we can grab it by subscribing to the subscription topic, but that would require a separate service just to log those payloads. "New subscriber" log entries are already there, so if `payload` can be logged as well it would be perfect.

Hopefully I don't miss anything from the docs that makes those points moot.

si14:
I think the bot was too eager to close it :(

dunglas:
Hi @si14,

Regarding `X-Forwarded-For`,  I think that this should be handled at Caddy's level. It's probably already to do this kind of rewrite (but I'm not sure if it's a good idea).

Regarding logging JWT `payload`, indeed it's a good idea, especially now that Caddy will allow to easily filter and redact log fields. Do you want to try to open a PR?

SherinBloemendaal:
How should i use it in Caddy? I've read something about the `reverse_proxy` directive that has a `trusted_proxies` option but i can't figure out how to use that in the mercure directive. What should i use for Caddy?

PR #779, opened 2023-06-01
feat(hub): add metadata to new subscriber log

aegypius:
Ref: #546

Add the content of the payload from jwt token when a new subscription occures.

### DIFF
--- a/subscribe.go
+++ b/subscribe.go
@@ -137,7 +137,7 @@ func (h *Hub) registerSubscriber(w http.ResponseWriter, r *http.Request) *Subscr
 	sendHeaders(w, s)
 
 	if c := h.logger.Check(zap.InfoLevel, "New subscriber"); c != nil {
-		if h.debug {
+		if h.logger.Level() == zap.DebugLevel {
 			c.Write(zap.Object("subscriber", s), zap.Reflect("payload", claims.Mercure.Payload))
 		} else {
 			c.Write(zap.Object("subscriber", s))
--- a/subscribe.go
+++ b/subscribe.go
@@ -96,9 +96,11 @@ func (h *Hub) registerSubscriber(w http.ResponseWriter, r *http.Request) *Subscr
 	s.Debug = h.debug
 	s.RemoteAddr = r.RemoteAddr
 	var privateTopics []string
+	var claims *claims
 
 	if h.subscriberJWT != nil {
-		claims, err := authorize(r, h.subscriberJWT, nil, h.cookieName)
+		var err error
+		claims, err = authorize(r, h.subscriberJWT, nil, h.cookieName)
 		if claims != nil {
 			s.Claims = claims
 			privateTopics = claims.Mercure.Subscribe
@@ -135,7 +137,7 @@ func (h *Hub) registerSubscriber(w http.ResponseWriter, r *http.Request) *Subscr
 	sendHeaders(w, s)
 
 	if c := h.logger.Check(zap.InfoLevel, "New subscriber"); c != nil {
-		c.Write(zap.Object("subscriber", s))
+		c.Write(zap.Object("subscriber", s), zap.Reflect("payload", claims.Mercure.Payload))
 	}
 	h.metrics.SubscriberConnected(s)
 
--- a/subscribe.go
+++ b/subscribe.go
@@ -137,11 +137,12 @@ func (h *Hub) registerSubscriber(w http.ResponseWriter, r *http.Request) *Subscr
 	sendHeaders(w, s)
 
 	if c := h.logger.Check(zap.InfoLevel, "New subscriber"); c != nil {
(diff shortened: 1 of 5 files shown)


dunglas:
Thanks for the PR!

I fear that the payload may contain sensitive data. Also, using `zap.Reflect()` will introduce a performance cost. Could be make this feature opt-in, or just enable it when the debug mode is on?

aegypius:
Thanks for the review, I will try to enable payload only if debug mode is on. 

Regarding `zap.Reflect` I am aware that this is probably not the most performant way but as far I know the payload is a free form JSON data so I can't implement a proper ObjectMarshaler (or maybe it's possible ?).

Sensitive data may be present in the payload but I don't know how to handle this without a proper schema to identify fields that may contain sensitive data.

subscribe_test.go
```diff
@@ -259,6 +259,44 @@ func TestSubscribe(t *testing.T) {
 	testSubscribe(t, 3)
 }
 
+func TestSubscribeWithDebug(t *testing.T) {
```
dunglas, line 262: You should assert that the payload has been logged using https://pkg.go.dev/go.uber.org/zap@v1.24.0/zaptest/observer

aegypius: @dunglas I added a test to check payload field is present in the expected log entry but I can't figure out how to pass and test the content of the payload.

dunglas: You need to create a JWT containing the data your expect: https://mercure.rocks/spec#payload

subscribe_test.go
```diff
@@ -259,6 +262,61 @@ func TestSubscribe(t *testing.T) {
...
+		for {
+			s.RLock()
+			empty := s.subscribers.Len() == 0
+			s.RUnlock()
+
+			if empty {
+				continue
+			}
+
+			hub.transport.Dispatch(&Update{
```
dunglas, line 279: Dispatching updates isn't necessary for this test as you log at subscription time.

dunglas (changes requested):

subscribe.go
```diff
@@ -135,7 +137,11 @@ func (h *Hub) registerSubscriber(w http.ResponseWriter, r *http.Request) *Subscr
 	sendHeaders(w, s)
 
 	if c := h.logger.Check(zap.InfoLevel, "New subscriber"); c != nil {
-		c.Write(zap.Object("subscriber", s))
+		if h.debug {
```
dunglas, line 140: IMHO you should check the debug level of the logger instead of relying on `h.debug`. This will allow user wanting to log detailed information in "prod" mode to do it.

You can use https://pkg.go.dev/go.uber.org/zap#Logger.Level to do the check.

subscribe_test.go
```diff
@@ -259,6 +263,51 @@ func TestSubscribe(t *testing.T) {
...
+	core, logs := observer.New(zapcore.DebugLevel)
+	var payload interface{}
+	json.Unmarshal([]byte("{\"bar\": \"baz\", \"foo\": \"bar\"}"), payload)
+
+	testSubscribeLogs(t, createDummy(
+		WithDebug(),
+		WithLogger(zap.New(core)),
+	), payload)
+
+	assert.True(t, logs.FilterMessage("New subscriber").FilterField(zap.Reflect("payload", payload)).Len() == 1)
```
dunglas, line 293: nitpicking:

```suggestion
	assert.Equal(t, 1, logs.FilterMessage("New subscriber").FilterField(zap.Reflect("payload", payload)).Len())
```

subscribe_test.go
```diff
@@ -259,6 +263,51 @@ func TestSubscribe(t *testing.T) {
...
+}
+
+func TestSubscribeWithoutDebug(t *testing.T) {
+	core, logs := observer.New(zapcore.DebugLevel)
+
+	var payload struct {
+		Bar string `json:"bar"`
+	}
+	payload.Bar = "baz"
```
dunglas, line 302: You can use inline initialization here:

```suggestion
	var payload struct {
		Bar string `json:"bar"`
	}{Bar: "baz"}
```

subscribe_test.go
```diff
@@ -259,6 +263,51 @@ func TestSubscribe(t *testing.T) {
...
+		cancel:             cancel,
+	}
+
+	hub.SubscribeHandler(w, req)
+}
+
+func TestSubscribeWithDebug(t *testing.T) {
+	core, logs := observer.New(zapcore.DebugLevel)
+	var payload interface{}
+	json.Unmarshal([]byte("{\"bar\": \"baz\", \"foo\": \"bar\"}"), payload)
```
dunglas, line 286: Initializing the struct "manually" doesn't work?

dunglas (approved):

dunglas:
Thank you @aegypius. I pushed a minor cleanup commit in your branch before merging.

PR #780, opened 2023-06-03
fix: wait for dispatch goroutine to finish

WeidiDeng:
Fix: subscribe handler only returns when the dispatch goroutine finishes.

As reported in [caddy](https://github.com/caddyserver/caddy/issues/5347#issuecomment-1573630081), it's not an error with caddy itself but with this plugin. [`http.ResponseWriter`](https://pkg.go.dev/net/http#ResponseWriter) is explicitly said may not be written to after the handler returns. This patch wait for the dispatch goroutine to finish then return when timed out.


WeidiDeng:
@dunglas I don't know why lint failed, seems to be like a bug with linter itself.

dunglas:
Thank you very much for the fix! I'll take care of the linters.

Issue #704, opened 2022-10-13
Mercure never catches you up when you ask for a last-event-id that it hasn't seen


spantaleev:
I have a system developed using an old alternative to Mercure, which I'd like to migrate to Mercure.

Clients ask for a version of a resource by sending `?lastEventID=..`.

These versions are simple timestamps (e.g. `1665666647`) stored on the resource (e.g. `Ticket.lastActivityTime = 1665666647`).

The problem is, Mercure does not necessarily get an update about each and every update. It only sees events about the important changes to this `lastActivityTime` field.

Because of this, sometimes clients will ask for a `lastEventID` which Mercure has never seen. Even though newer updates are published to Mercure, the client that asked for `?lastEventID=OLD_AND_UNSEEN` will never get updated. It will be stuck forever.

---

I was hitting the same issue with clients asking for `lastEventID=0` (when they haven't seen any updates) yet. I've worked around this by sending `lastEventID=earliest` as the spec says.

However, I still have some instances in which clients ask for a missing event ID. These clients neither get an error, nor a response telling them they're in the wrong, nor do they magically get caught up somehow (I realize that there's no good way to magically catch them up).

dunglas:
This isn't.

h0jeZvgoxFepBQ2C:
Whops sorry, my fault! 

Btw: would it be possible to get some feedback on this issue? We are evaluating mercure and are browsing through the issues to get some feeling about the current state / issues. Thank you so much for your great work ❤️

slim:
I tested this. Current behaviour is to ignore any bogus ID and serve the updates as if `&lastEventID` was not requested. Here's the relevant code : 

https://github.com/dunglas/mercure/blob/main/bolt_transport.go#L234

Not sure what you mean by "the client that asked for ?lastEventID=OLD_AND_UNSEEN will never get updated. It will be stuck forever.". So not sure if the bug is still there.

Also not sure if this is the correct behaviour, the [Spec](https://mercure.rocks/spec#reconciliation) does not address the case where bogus `&lastEventID` is given, but does say that client should not rely on Hub answering the request for specific subset of history when it says "The reserved value [earliest] can be used to hint the hub to send all updates it has for the subscribed topics. According to its own policy, the hub MAY or MAY NOT fulfil this request."

PR #782, opened 2023-06-08
feat: log bogus LastEventID when using BoltDB

slim:
Logging Bogus LastEventID given by client

Related to: #704

### DIFF
--- a/bolt_transport.go
+++ b/bolt_transport.go
@@ -257,7 +257,7 @@ func (t *BoltTransport) dispatchHistory(s *Subscriber, toSeq uint64) {
 		s.HistoryDispatched(responseLastEventID)
 		if !afterFromID {
 			if c := t.logger.Check(zap.ErrorLevel, "Can't find LastEventID in Bolt DB"); c != nil {
-				c.Write(zap.Error(ErrIDNotFound))
+				c.Write(zap.Error(fmt.Errorf("id not found %s", s.RequestLastEventID)))
 			}
 		}
 
--- a/bolt_transport.go
+++ b/bolt_transport.go
@@ -255,6 +255,7 @@ func (t *BoltTransport) dispatchHistory(s *Subscriber, toSeq uint64) {
 			}
 		}
 		s.HistoryDispatched(responseLastEventID)
+		//nolint:gocognit
 		if !afterFromID {
 			if c := t.logger.Check(zap.InfoLevel, "Can't find requested LastEventID"); c != nil {
 				c.Write(zap.String("LastEventID", s.RequestLastEventID))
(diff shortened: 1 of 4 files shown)


bolt_transport.go
```diff
@@ -255,6 +255,11 @@ func (t *BoltTransport) dispatchHistory(s *Subscriber, toSeq uint64) {
 			}
 		}
 		s.HistoryDispatched(responseLastEventID)
+		if !afterFromID {
+			if c := t.logger.Check(zap.ErrorLevel, "Can't find LastEventID in Bolt DB"); c != nil {
+				c.Write(zap.Error(ErrIdNotFound))
```
dunglas, line 260: This isn't necessary to create an error type as we don't use it. You can log a string directly.

dunglas (changes requested):

bolt_transport.go
```diff
@@ -255,6 +255,11 @@ func (t *BoltTransport) dispatchHistory(s *Subscriber, toSeq uint64) {
 			}
 		}
 		s.HistoryDispatched(responseLastEventID)
+		if !afterFromID {
+			if c := t.logger.Check(zap.ErrorLevel, "Can't find LastEventID in Bolt DB"); c != nil {
+				c.Write(zap.Error(fmt.Errorf("id not found %s", s.RequestLastEventID)))
```
dunglas, line 260:
```suggestion
				c.Write(zap.String("LastEventID", s.RequestLastEventID)))
```

No need to log the Go error type here, just logging the ID in a structured field will be good enough.

bolt_transport.go
```diff
@@ -255,6 +255,11 @@ func (t *BoltTransport) dispatchHistory(s *Subscriber, toSeq uint64) {
 			}
 		}
 		s.HistoryDispatched(responseLastEventID)
+		if !afterFromID {
+			if c := t.logger.Check(zap.ErrorLevel, "Can't find LastEventID in Bolt DB"); c != nil {
```
dunglas, line 259:
```suggestion
			if c := t.logger.Check(zap.ErrorLevel, "Can't find requested LastEventID"); c != nil {
```

No need to mention Bolt DB, so we can log the exact same messages for other transports and ease switch between transports.

dunglas: Also, the `Info` level will be good enough, WDYT? (Error would mean waking up ops just for a client providing a bad ID).

dunglas (changes requested):

dunglas (approved):
Could you also add a test and fix the linter please?

dunglas:
Thank you!

PR #792, opened 2023-08-01
fix(helm): topologySpreadConstraints example value

dunglas:
According to my tests, an indentation level is missing in https://github.com/dunglas/mercure/pull/730.
Could you check if it looks ok to you @jfcoz, please?

Also, is it a common practice to include Helm template code as a string in values? The NGINX Ingress is using a plain array instead for instance. WDYT?
This may also be why this value isn't currently automatically documented by `helm-docs` (or because we need to specify this key with a default value).

### DIFF
--- a/charts/mercure/values.yaml
+++ b/charts/mercure/values.yaml
@@ -144,7 +144,7 @@ affinity: {}
 #    whenUnsatisfiable: DoNotSchedule
 #    labelSelector:
 #      matchLabels:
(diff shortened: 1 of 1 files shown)


jfcoz:
Yes for nindent 8.
For the template code, in nginx the labelSelector is hardcoded here : https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L250-L256
with the template, and the tpl function usage, it allows to pass a value which will depend of each release name.

PR #793, opened 2023-08-02
feat: add PodDisruptionBudget

jfcoz:
add PodDisruptionBudget when replicaCount > 1 or autoscaling is enabled with minReplica > 1


dunglas:
Shouldn't this change be done in the open source repo?

charts/mercure/Chart.yaml
```diff
@@ -12,7 +12,7 @@ keywords:
 - streaming-api
 - async-api
 type: application
-version: 0.14.10
+version: 0.14.11
```
dunglas, line 15: To revert, this is handled automatically by the release script.

Issue #771, opened 2023-05-24
Feature request: keep publisherJwtKey subscriberJwtKey in external k8s secret


warmanton:
Hello.
This is feature request about keeping publisherJwtKey and subscriberJwtKey in external k8s secret.
Now I have to specify these variables via values of chart. There is no an option to specify external existing secret for these jwtKeys.

Usually helm chars have an option for the same cases: user can specify some secret directly via values (less secure) or user can configure a helm release to use existing secret (more secure)

PR #809, opened 2023-09-18
Allow passing existing secret in helm chart values

Julian-Louis:
Hello 👋
This pull request will allow passing an existing secret name into the helm chart values.
This is useful when using the GitOps approach, where values need to be committed to a git repository.

**Related issues:**
- https://github.com/dunglas/mercure/issues/771

### DIFF
--- a/charts/mercure/templates/_helpers.tpl
+++ b/charts/mercure/templates/_helpers.tpl
@@ -60,3 +60,14 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Get the secret name.
+*/}}
+{{- define "mercure.secretName" -}}
+{{- if .Values.existingSecret -}}
+{{- printf "%s" (tpl .Values.existingSecret $) -}}
+{{- else -}}
+{{- printf "%s" (include "mercure.fullname" .) -}}
+{{- end -}}
+{{- end -}}
--- a/charts/mercure/templates/deployment.yaml
+++ b/charts/mercure/templates/deployment.yaml
@@ -50,17 +50,17 @@ spec:
             - name: EXTRA_DIRECTIVES
               valueFrom:
                 secretKeyRef:
-                  name: {{ include "mercure.fullname" . }}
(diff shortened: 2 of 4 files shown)


VincentDugard:
Hello, 
why it has been closed ? as stated this would be usefull to work with tools like eso

Julian-Louis:
Hello @VincentDugard 
I didn't get any feedback, so I figured my pr wasn't very useful for this project.

dunglas:
@Julian-Louis do you have any example of an "official" (or Bitnami) chart, or a link pointing to the relevant documentation entry? It looks weird to me to allow to entirely bypass the values system.

VincentDugard:
@dunglas There are multiple bitnami charts that use this system. For exemple in [the postgresql chart ](https://artifacthub.io/packages/helm/bitnami/postgresql#postgresql-common-parameters), you can provide an existing secrets for postgres credentials. There is also the possibility to override key name in the secret.

dunglas:
Indeed, to me, it makes more sense to allow this for the JWTs!

jnunez-clicars:
Why not to just allow to add an extra secret to the env? So the deployment needs only a new `envFrom` at the end and the user can create a secret with only the two subscriberJWT and publisherJWT keys

```yaml
          env:
            # keep env as is
            # ...
            - name: MERCURE_PUBLISHER_JWT_ALG
              valueFrom:
                configMapKeyRef:
                  name: {{ include "mercure.fullname" . }}
                  key: publisher-jwt-alg
            # ...
          # allow to overwrite config
          envFrom:
            - secretRef:
                name: the-user-secret-with-the-keys-he-wants
```

lerminou:
Hi @dunglas, Do you have any motivation to not merge it ? 
We also want to keep the same JWT key between deployments with argocd, with our external-secret system.

It would allow to use an existing secret in our cluster. A must have feature.

dunglas (approved):

dunglas:
Thanks!

janosmiko:
Hi @dunglas ,

Could you prepare a new release on the helm chart to adapt this change?

PR #813, opened 2023-10-01
adding support for JWKS

broncha:
Adds support for JWKS in Mercure. With this update, you would simply configure the JWKS URL and Mercure would validate the subscriber and publisher JWT based on the Key ID and the keys in the JWKS.
Still a work in progress, as most tests that look for JWT need to be duplicated to test JWKS

### DIFF
--- a/authorization.go
+++ b/authorization.go
@@ -1,11 +1,13 @@
 package mercure
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 
+	"github.com/MicahParks/keyfunc"
 	"github.com/golang-jwt/jwt/v4"
 	"go.uber.org/zap"
 )
@@ -48,18 +50,20 @@ var (
 	ErrInvalidJWT = errors.New("invalid JWT")
 	// ErrPublicKey is returned when there is an error with the public key.
 	ErrPublicKey = errors.New("public key error")
+	// ErrInvalidJWKS is returned when the JWKS config is invalid.
+	ErrInvalidJWKS = errors.New("invalid JWKS")
 )
 
 // Authorize validates the JWT that may be provided through an "Authorization" HTTP header or an authorization cookie.
 // It returns the claims contained in the token if it exists and is valid, nil if no token is provided (anonymous mode), and an error if the token is not valid.
-func authorize(r *http.Request, jwtConfig *jwtConfig, publishOrigins []string, cookieName string) (*claims, error) {
+func authorize(r *http.Request, jwtConfig *jwtConfig, jwks *jwksConfig, publishOrigins []string, cookieName string) (*claims, error) {
 	authorizationHeaders, headerExists := r.Header["Authorization"]
 	if headerExists {
 		if len(authorizationHeaders) != 1 || len(authorizationHeaders[0]) < 48 || authorizationHeaders[0][:7] != bearerPrefix {
 			return nil, ErrInvalidAuthorizationHeader
 		}
 
-		return validateJWT(authorizationHeaders[0][7:], jwtConfig)
+		return validateJWT(authorizationHeaders[0][7:], jwtConfig, jwks)
 	}
 
 	if authorizationQuery, queryExists := r.URL.Query()["authorization"]; queryExists {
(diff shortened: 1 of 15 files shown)


authorization.go
```diff
@@ -98,30 +99,43 @@ func authorize(r *http.Request, jwtConfig *jwtConfig, publishOrigins []string, c
...
-		case *jwt.SigningMethodHMAC:
-			return jwtConfig.key, nil
-		case *jwt.SigningMethodRSA:
-			pub, err := jwt.ParseRSAPublicKeyFromPEM(jwtConfig.key)
-			if err != nil {
-				return nil, fmt.Errorf("unable to parse RSA public key: %w", err)
+func validateJWT(encodedToken string, jwtConfig *jwtConfig, jwksURL string) (*claims, error) {
+	var keyFunc jwt.Keyfunc
+
+	if jwksURL != "" {
```
dunglas, line 113: nitpicking: I would invert the test to avoid the negation as there is an `else`.

authorization.go
```diff
@@ -98,30 +99,43 @@ func authorize(r *http.Request, jwtConfig *jwtConfig, publishOrigins []string, c
...
+		jwks, err := keyfunc.Get(jwksURL, keyfunc.Options{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to get the JWKS from the given URL.\nError: %w", err)
```
dunglas, line 116:
```suggestion
			return nil, fmt.Errorf("failed to get the JWKS from the given URL: %w", err)
```

authorization.go
```diff
@@ -98,30 +99,43 @@ func authorize(r *http.Request, jwtConfig *jwtConfig, publishOrigins []string, c
...
+		}
+
+		keyFunc = jwks.Keyfunc
+	} else {
+		keyFunc = func(token *jwt.Token) (interface{}, error) {
```
dunglas, line 121: Maybe could you use a real symbol instead of an anonymous function to ease testing?

hub.go
```diff
@@ -146,6 +146,22 @@ func WithSubscriberJWT(key []byte, alg string) Option {
 	}
 }
 
+func WithSubscriberJWKS(jwks string) Option {
```
dunglas, line 149: It could be cool to support all methods supported by the underlying lib (URL, JSON, and raw key). WDYT?

dunglas (changes requested):
Thank you very much for working on this, this is a very needed feature.

hub.go
broncha: As far as I have worked with JWKS, I have always worked with a URL. I think using JSON and raw key defeats the purpose of using the JWKS support as you would need to update those and restart the services in the event of key rotation. I would like to support the `keyfunc.Options` though. What would you suggest for those options in Caddy config? I was thinking of a nested block.

authorization.go
broncha: The reason I did that is to prioritize jwks when both jwks and jwt config are present. Do you see a way we could combine these?

hub.go
dunglas: Nested blocks looks great. I'm pretty sure that some advanced users will want to use other options so we should support both yet. For instance, advanced CI/CD pipelines may want to bundle the trusted keys.

broncha:
@dunglas Finallly had some time to come back to this again. It now supports URL, json or key as jwks config.

dvv:
Please consider applying.

dunglas:
We cannot merge the path as-is because it uses the deprecated version of `keyfunc` and `jwt`.
I rebased and slightly adjusted this patch (https://github.com/dunglas/mercure/compare/jwks_support) but there is more work to do.

broncha:
I might have some time to work on #851 in coming weeks.

dunglas:
@broncha just rebasing on top of #822 and upgrading to the latest version of `keyfunc` should be enough. The latest version seems to provide a simplified API, so we just need a new option to provide the server JWKS server URL and that should be enough.

dunglas:
@broncha btw, do you mind if I force-push my rebase in your branch?

broncha:
@dunglas Please go ahead. Ill pick it up from there.

broncha:
I would also like to unify the jwt and jwks configs too. But I am not sure about the current status of the configs. Ill check those when I get a change to look at this.

broncha:
Also for the record, one issue have been repeatedly facing is [clock skew](https://github.com/golang-jwt/jwt/issues/98), where jwt creation and validation happens almost at the same time. It would be good to be able to support, which landed in golang-jwt v5.

PR #824, opened 2023-10-13
docs: remove superfluous `xcaddy` flag

mohammed90:
The Go module `github.com/dunglas/mercure/caddy` imports the Go module `github.com/dunglas/mercure`, which means it's already part of the build.


dunglas:
Thanks!

Issue #588, opened 2021-11-17
Switch from URI Templates to URI Pattern


dunglas:
There is a new proposal for a standard dedicated to matching URLs: https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API

If it is adopted, it will better fit the Mercure use case than how we currently hijack the URI Template spec.

dunglas:
We currently lack a Go library implementing the URLPattern spec.

Some useful resources to implement this:

* [Test data](https://github.com/web-platform-tests/wpt/tree/master/urlpattern/resources
) used by Chrome, Deno, and the JS Polyfill
* [Rust implementation](https://github.com/denoland/rust-urlpattern)
* [JS implementation](https://github.com/kenchris/urlpattern-polyfill)
* [C++ library used by Chrome](https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/liburlpattern0)

Similar code that may be adapted to conform to the spec:

* [path-to-regexp (Go port)](https://github.com/soongo/path-to-regexp)
* [Go Fiber's path](https://github.com/gofiber/fiber/blob/master/path.go)

PR #831, opened 2023-11-03
spec: use URL patterns instead of URI templates

dunglas:
The [URL Pattern standard](https://urlpattern.spec.whatwg.org) has finally been published. This is great news for Mercure, as it was the only missing part to publish a definitive version of the protocol, and a version 1.0 of the hub.

Until now, we've allowed multiple resources to be selected to subscribe to notifications using URI patterns (RFC 6570). This approach works, but has a few drawbacks:

1. As stated in the [URI Templates](https://datatracker.ietf.org/doc/html/rfc6570) (RFC 6570) specification, URI templates are not designed for matching:
>    Some URI Templates can be used in reverse for the purpose of variable matching: comparing the template to a fully formed URI in order to extract the variable parts from that URI and assign them to the named variables.  Variable matching only works well if the template expressions are delimited by the beginning or end of the URI or by  characters that cannot be part of the expansion, such as reserved characters surrounding a simple string expression.  In general, regular expression languages are better suited for variable matching.
2. Matching with URI Templates is more limited than with regular expressions.
3. The syntax of URI Templates, especially when used for matching, is not very popular and most web developers have difficulty mastering it. This is a problem regularly encountered by Mercury users. We've even created a tool to help them: https://uri-template-tester.mercure.rocks. Regular expression syntax is much more popular.

This (hopefully last) change in the specification replaces usage or URI Templates by URL patterns for subscribing to several topics at the same time.

I've also taken the opportunity to separate the comparison of raw topics (exact comparison) from the comparison with URL patterns. The `topic` query parameter is still used for raw string comparison, but a new `pattern` query parameter is used for URL templates. In addition, support for pattern comparison by hubs is now optional, facilitating protocol implementation.
This change simplifies the implementation of the specification, and makes it possible to switch from URI patterns to URL patterns (a hub will be able to support both obsolete patterns and new ones). This is important, as Mercury is now very popular.

@wanderview, as the lead author of the URL Pattern standard, I wonder if you could verify that we are using the standard as intended.

For #588.

Next steps:

* [X] write an implementation of URL Pattern in Go (work in progress) (done: https://github.com/dunglas/go-urlpattern)
* [ ] patch the hub to implement this change using this new library

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ If you include code from another project, please mention it in the Pull Request 
 
 ## Commit Messages
 
-The commit message must follow the [Conventional Commits specification](https://www.conventionalcommits.org/).
+The commit message must follow the [Conventional Commits specification](https://www.conventionalcommits.org).
 The following types are allowed:
 
 * `fix`: bugfix
--- a/spec/mercure.md
+++ b/spec/mercure.md
@@ -45,19 +45,19 @@ NOT**, **RECOMMENDED**, **MAY**, and **OPTIONAL**, when they appear in this docu
 interpreted as described in [@!RFC2119].
 
 *   Topic: The unit to which one can subscribe to changes. The topic **SHOULD** be identified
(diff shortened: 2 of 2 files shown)


dunglas:
Closing in favor of #959.

PR #868, opened 2024-01-30
fix: Mercure container port can be modified

Kadeux:
Due to [Openshift requirements](https://docs.openshift.com/acs/3.72/operating/default-security-policies.html), containers can not be exposed on ports < 1024. The mercure container is mapped on port 80.


dunglas (approved):
LGTM

slim (approved):

dunglas:
Thank you!

PR #885, opened 2024-03-19
feat: add suport JSON Web Key Set

dunglas:
Close #813.

I haven't tested this patch in real life yet. Could someone tell me if it works as expected (@broncha maybe)?


broncha:
@dunglas tested with the jwks URLs and it all works as expected!

dunglas:
Thanks for the feedback.

PR #888, opened 2024-04-05
docs: env var renamed in Caddyfile

maxhelias:
EXTRA_DIRECTIVES has been renamed CADDY_SERVER_EXTRA_DIRECTIVES


dunglas (approved):

dunglas:
Good catch! Thanks!

PR #924, opened 2024-07-04
feat: Add support for metrics in chart

coibib:
This PR adds a specific port for the metrics in the Helm Chart.

### DIFF
--- a/charts/mercure/templates/service.yaml
+++ b/charts/mercure/templates/service.yaml
@@ -18,3 +18,27 @@ spec:
       name: http
   selector:
     {{- include "mercure.selectorLabels" . | nindent 4 }}
+---
+{{- if .Values.metrics.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "mercure.fullname" . }}-metrics
+  labels:
+    {{- include "mercure.labels" . | nindent 4 }}
+    caca: proute
+  {{- with .Values.service.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+
+spec:
+  type: ClusterIP
+  ports:
+    - port: {{ .Values.metrics.port }}
+      targetPort: metrics
+      protocol: TCP
+      name: metrics
+  selector:
+    {{- include "mercure.selectorLabels" . | nindent 4 }}
+{{- end }}
--- a/charts/mercure/templates/service.yaml
+++ b/charts/mercure/templates/service.yaml
@@ -26,7 +26,7 @@ metadata:
   name: {{ include "mercure.fullname" . }}-metrics
   labels:
     {{- include "mercure.labels" . | nindent 4 }}
-    prometheus-operated: true
+    app.kubernetes.io/component: metrics
   {{- with .Values.service.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
--- a/charts/mercure/templates/service.yaml
+++ b/charts/mercure/templates/service.yaml
@@ -26,7 +26,7 @@ metadata:
   name: {{ include "mercure.fullname" . }}-metrics
   labels:
     {{- include "mercure.labels" . | nindent 4 }}
-    caca: proute
+    prometheus-operated: true
   {{- with .Values.service.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
--- a/charts/mercure/templates/serviceMonitor.yaml
+++ b/charts/mercure/templates/serviceMonitor.yaml
@@ -4,12 +4,25 @@ kind: ServiceMonitor
 metadata:
   name: {{ include "mercure.fullname" . }}-metrics
   labels:
-    app.kubernetes.io/name: {{ include "mercure.name" . }}
+    {{- include "mercure.labels" . | nindent 4 }}
+    app.kubernetes.io/component: metrics
 spec:
   selector:
     matchLabels:
-      prometheus-operated: "true"
+      app.kubernetes.io/component: metrics
   endpoints:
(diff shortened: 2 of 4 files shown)


xvilo:
Do we also want to add support for prometheus service monitor? https://observability.thomasriley.co.uk/prometheus/configuring-prometheus/using-service-monitors/ 

In this case it would be automatically be picked up by prometheus-operator running on my cluster :D

Kadeux:
> Do we also want to add support for prometheus service monitor? https://observability.thomasriley.co.uk/prometheus/configuring-prometheus/using-service-monitors/
> 
> In this case it would be automatically be picked up by prometheus-operator running on my cluster :D

I totally agree with you, I added the corresponding changes.

xvilo (approved):

xvilo:
LGTM

charts/mercure/templates/service.yaml
```diff
@@ -18,3 +18,27 @@ spec:
...
     {{- include "mercure.selectorLabels" . | nindent 4 }}
+---
+{{- if .Values.metrics.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "mercure.fullname" . }}-metrics
+  labels:
+    {{- include "mercure.labels" . | nindent 4 }}
+    prometheus-operated: true
```
dunglas, line 29:
```suggestion
    prometheus-operated: true
    app.kubernetes.io/component: metrics
```

dunglas, line 29: Why is this label necessary? I don't find where it is documented.

charts/mercure/templates/serviceMonitor.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if and .Values.metrics.enabled .Values.metrics.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "mercure.fullname" . }}-metrics
+  labels:
+    app.kubernetes.io/name: {{ include "mercure.name" . }}
```
dunglas, line 7:
```suggestion
  labels:
    {{- include "mercure.labels" . | nindent 4 }}
    app.kubernetes.io/component: metrics
```

charts/mercure/templates/serviceMonitor.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  selector:
+    matchLabels:
+      prometheus-operated: "true"
```
dunglas, line 11:
```suggestion
      app.kubernetes.io/component: metrics
```

charts/mercure/templates/serviceMonitor.yaml
```diff
@@ -0,0 +1,15 @@
...
+  endpoints:
```
dunglas, line 12: We should likely include options similar to these ones too:  https://github.com/bitnami/charts/blob/1dccc11c8e8cfa80aad6fc6628d1edc02c49b4a4/bitnami/wordpress/templates/servicemonitor.yaml#L22-L36

charts/mercure/values.yaml
```diff
@@ -28,6 +28,22 @@ subscriberJwtAlg: HS256
 # -- The license key for [the High Availability version](https://mercure.rocks/docs/hub/cluster) (not necessary is you use the FOSS version).
 license: ""
 
+metrics:
+  # -- Enable the metrics. You must also add the [metrics](https://caddyserver.com/docs/caddyfile/directives/metrics) in the `globalOptions` for Caddy server.
+  # servers {
```
dunglas, line 33: Can't we add this option automatically using the by appending the needed config to the `GLOBAL_OPTIONS` environment variable in `deployment.yaml`?

dunglas (changes requested):
Thanks for working on this, and sorry for the late review.

I made some small suggestions inspired by https://github.com/bitnami/charts/blob/main/bitnami/wordpress/.

gabrielrosset:
Hello @dunglas , it seems that @coibib has taken your remarks into account, is there anything left to do on this PR to merge it?

Hello @coibib , can you please allow me to push on your fork to allow us to get the subject back into the project ? Thanks.

charts/mercure/values.yaml
dunglas: Indeed, let's do that like this for now. We could do something smarter in the future (like detection if `servers` is not part of the existing string, but we'll see in follow-up PRs!

charts/mercure/values.yaml
```diff
@@ -108,7 +134,8 @@ ingress:
 # -- Container resource [requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/).
 # See the [API reference](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#resources) for details.
 # @default -- No requests or limits.
-resources: {}
+resources:
+  {}
```
dunglas, line 138: Why this change?

dunglas (approved):

charts/mercure/values.yaml
coibib: This is my linter, i reverted the change back.

xvilo:
Hiya, whats the status of this PR? We've been using Mercure in production since a month or so, and running it through kubernetes. Unfortunately we've had issues with it borking twice now, and with just the ingress logs we don't have enough data :-) so this might be super helpful 👍

dunglas (approved):

dunglas:
Thanks @coibib!

dunglas:
@xvilo merged, it will be in the next release.

PR #937, opened 2024-08-18
feat(caddy): allow passing the context to the transport

dunglas:
A better solution would be to create a Caddy module per transport.


dunglas:
Closing on favor of #939.

PR #939, opened 2024-08-19
feat!: use Caddy modules for transports

dunglas:
**This patch introduces BC breaks described in the `UPGRADE` file**

This patch deprecates the `transport_url` and introduces Caddy modules dedicated to each transport.

This approach has several benefits:

* cleaner configuration: it's now possible to use the full expressivity of the `Caddyfile` syntax instead of relying on embedded URLs in the config
* improved extensibility: transports (including built-in transports, custom transports, and transports provided by the paid version) can now expose a structured config and access config from the main Mercure Caddy module, which was complex before
* improved reliability: transports can hook into Caddy lifecycle events (config change, shutdown...). That wasn't previously possible

### DIFF
--- a/caddy/bolt_transport.go
+++ b/caddy/bolt_transport.go
@@ -1,0 +1,128 @@
+package caddy
+
+import (
+	"bytes"
+	"encoding/gob"
+	"strconv"
+
+	"github.com/caddyserver/caddy/v2"
+	"github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+	"github.com/dunglas/mercure"
+)
+
+func init() { //nolint:gochecknoinits
+	caddy.RegisterModule(Bolt{})
+}
+
+type Bolt struct {
+	Path             string  `json:"path,omitempty"`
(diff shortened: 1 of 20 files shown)


caddy/bolt_transport.go
```diff
@@ -0,0 +1,128 @@
...
+	CleanupFrequency float64 `json:"cleanup_frequency,omitempty"`
+
+	transport    *mercure.BoltTransport
+	transportKey string
+}
+
+// CaddyModule returns the Caddy module information.
+func (Bolt) CaddyModule() caddy.ModuleInfo {
+	return caddy.ModuleInfo{
+		ID:  "http.handlers.mercure.bolt",
```
francislavoie, line 30: Hopefully it's not too late, I would recommend grouping these modules under a namespace like `http.handlers.mercure.transports.*` because as-is if you ever want to modularize something else within mercure then the entire namespace under it is already taken by the transports. So if you make another namespace like `http.handlers.mercure.whatever.*` technically that's inside of `http.handlers.mercure.*` which you're already using now for transports.

PR #941, opened 2024-08-21
docs: Update docker compose healthcheck

dallegoet:
I replaced `curl` by `wget` in the healthcheck of `docker-compose.yml`, because curl is not available in the image `dunglas/mercure`.

The arguments :
- `-q` : quiet
- `--spider` : don't download anything

If there is no error `wget` returns 0, else 1


dunglas (approved):

dunglas:
Thanks!

PR #946, opened 2024-08-31
spec!: make payloads topic-specific

dunglas:
Implementation: #945

### DIFF
--- a/docs/UPGRADE.md
+++ b/docs/UPGRADE.md
@@ -1,6 +1,32 @@
 # Upgrade
 
-## 0.17
+## 0.20
+
+The `mercure.payload` JWT key has been deprecated. It is now possible to make topic-specific data
+available in subscriptions events and through the subscription API.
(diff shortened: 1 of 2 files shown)


docs/UPGRADE.md
```diff
@@ -1,6 +1,32 @@
...
+}
+```
+
+After:
+
+```json
+{
+  "mercure": {
+    "payloads": {
+      "*": {"foo": "bar"}
```
soyuka, line 25: can you maybe add an example with a topic such as 
```
{
    "subscribe": ["https://example.com/foo", "https://example.com/bar/baz"]
    "payloads": {
        "https://example.com/bar/{val}": {"custom": "data only available for subscriptions matching this selector"},
        "*": {"data": "available for all other topics"}
    }
}
```

soyuka: or link the doc page maybe?

PR #949, opened 2024-09-14
feat: support ECDSA and Ed25519 JWT public keys

erkia:
Some more modern signing algorithms were missing from Mercure.

Also, refactored the authentication tests so more algorithms could be easily tested without copy-pasting bunch of functions.


dunglas (approved):

dunglas:
Thank you very much. That's great to see such high-quality PRs ❤️

PR #962, opened 2024-10-22
feat(helm): Make nodePort configurable for NodePort service type

bebehei:
When using `NodePort` as the `service.type` in the helm chart, there is no possibility to fix the port number.

This introduces the conditional for the helm chart.


dunglas (approved):

dunglas:
Thank you

PR #967, opened 2024-11-04
fix: support autoscaling/v2

xip-niekneuij:
autoscaling/v2beta1 isn't supported on Kubernetes v1.25.

See:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/


dunglas (approved):

dunglas:
Thank you!

PR #968, opened 2024-11-05
fix(helm): support autoscaling/v2 formatting

xip-niekneuij:
In #967, I forgot to add the new formatting for autoscaling/v2, which are different from autoscaling/v2beta1

Please accept my apologies for this oversight.


dunglas:
No worries! Thank you for the fix.

PR #974, opened 2024-11-07
fix: Update dev.Caddyfile: Fixing indentation

ThomasLandauer:
> WARN    Caddyfile input is not formatted; run 'caddy fmt --overwrite' to fix inconsistencies


dunglas:
Thank you, finished in #978.

PR #993, opened 2024-12-10
docs: Mention dev.Caddyfile file change in docs

paulinevos:
Users on Docker images older than v0.16.2 may be confused when they run the container, as the file `dev.Caddyfile` doesn't exist on those containers (v0.16.2 changed it from `Caddyfile.dev` to `dev.Caddyfile`)


dunglas (approved):

paulinevos:
@dunglas I think something is wrong with the CLA check, because I signed it a few hours ago

dunglas:
Thank you @paulinevos. That's weird for the CLA.

Issue #986, opened 2024-11-21
Shutting down Mercure server takes indefinitely  


bobvandevijver:
I was looking into a very slow shutdown of the Mercure instance I am running on a Ubuntu server, using a systemd unit. 

<details>
<summary>Unit file</summary>

```
[Unit]
Description=Mercure server
After=network.target

[Service]
Type=simple
EnvironmentFile=/opt/application/mercure/env
ExecStart=/opt/application/mercure/mercure run
Restart=always
RestartSec=5
RuntimeMaxSec=7d
WorkingDirectory=/opt/application/mercure
User=www-data

[Install]
WantedBy=multi-user.target
```

</details>

<details>
<summary>Logs</summary>

```
Nov 21 20:15:40 linux645 mercure[1197]: {"level":"info","ts":1732216540.629172,"msg":"shutting down apps, then terminating","signal":"SIGTERM"}
Nov 21 20:15:40 linux645 mercure[1197]: {"level":"warn","ts":1732216540.6292357,"msg":"exiting; byeee!! �<9F><91><8B>","signal":"SIGTERM"}
Nov 21 20:15:40 linux645 mercure[1197]: {"level":"info","ts":1732216540.6292503,"logger":"http","msg":"servers shutting down with eternal grace period"}
Nov 21 20:15:40 linux645 systemd[1]: Stopping mercure.service - Mercure server...
Nov 21 20:16:19 linux645 mercure[1197]: {"level":"info","ts":1732216579.8925092,"logger":"http.handlers.mercure","msg":"Subscriber disconnected","subscriber":{"id":"urn:uuid:adce74f5-1f27-4912-91b6-83cd4442f0a>
Nov 21 20:16:19 linux645 mercure[1197]: {"level":"info","ts":1732216579.89257,"logger":"http.log.access.log0","msg":"handled request","request":{"remote_ip":"127.0.0.1","remote_port":"55950","client_ip":"127.0>
Nov 21 20:16:31 linux645 mercure[1197]: {"level":"info","ts":1732216591.1713917,"logger":"http.handlers.mercure","msg":"Subscriber disconnected","subscriber":{"id":"urn:uuid:5db2a139-2a36-4f4d-a7e9-e4b1ed7d6fb>
[...]

Issue #857, opened 2024-01-17
Upstream Error when Using Mercure with Nginx Ingress on Kubernetes


alpkahvecioglu:
I am trying to install Mercure with the provided Helm Chart with the ingress enabled. Our ingress controller is nginx-ingress so that the traffic is coming to Mercure from Nginx. However, requests fail with **upstream prematurely closed connection while reading upstream**.

I tried configuring the nginx side with the following  [configuration](https://mercure.rocks/docs/hub/nginx) but the error still happens.

I am not sure how to set **use_forwarded_headers** with the Caddy installation of Mercure.

Any help is appreciated.

dunglas:
FIxed by #1005.

Issue #854, opened 2024-01-03
Help for configuring Mercure behind nginx as proxy-server


Astro-Otter-Space:
Hello,

I'm deploying Mercure hub on my server (VPS hosted by OVH) where an API REST (Symfony+FOSRestBundle) and a front app (VueJS 3, vue-cli) are installed yet, served with Nginx. API and FO have their own domains (https://api.exemple.space/ and https://www.exemple.space/). I've created a domain for Mercure (https://mercure.exemple.com/). Mercure (version 0.15) is installed with binary, not with docker.

I  added an nginx host configuration for working as reverse-proxy (https://mercure.rocks/docs/hub/nginx):
```php
# should i keep these lines commented or not ?
#server {
#  listen      80 http2;
#  server_name mercure.exemple.space;
#  return 301 https://mercure.exemple.space;
#}

server {
  listen      443 ssl http2;
  listen [::]:443 ssl http2;
  server_name mercure.exemple.space;

  ssl_certificate /etc/letsencrypt/live/mercure.exemple.space/fullchain.pem; 
  ssl_certificate_key /etc/letsencrypt/live/mercure.exemple.space/privkey.pem;
  
  location / {
    proxy_pass http://127.0.0.1:3000; # <-- is correct ?
    proxy_read_timeout 24h;
    proxy_http_version 1.1;
    proxy_set_header Connection "";
    proxy_connect_timeout 300s;

    #proxy_set_header Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-Proto $scheme;
  }

  # Configuration des logs
  access_log  /var/log/nginx/mercure_access.log;
  error_log   /var/log/nginx/mercure_error.log;
}
```
 i set public and private keys and set them as env variables in /etc/profile.d/mercure.sh
```bash
export MERCURE_PUBLISHER_JWT_KEY=$(cat /home/me/mercure/publisher.key.pub)
export MERCURE_PUBLISHER_JWT_ALG=RS256
export MERCURE_SUBSCRIBER_JWT_KEY=$(cat /home/me/mercure/subscriber.key.pub)
export MERCURE_PUBLISHER_JWT_ALG=RS256
export SERVER_NAME=localhost:3000 # <--  mercure.exemple.space or mercure.exemple.space:3000, which one is correct ?
```

[...]

Issue #843, opened 2023-12-07
Mercure inside docker and behind server host Nginx proxy


kl3sk:
Hello,

I faced a problem. A create a route to test my update and here is the message

```
fopen(): SSL operation failed with code 1. OpenSSL Error messages:
error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed
```

Docker config
```yaml
networks:
    web-staging-network:
        driver: bridge

services:
    mercure:
        container_name: mercure
        networks:
            web-staging-network:
        env_file:
            - ./.env
        image: dunglas/mercure:latest
        restart: unless-stopped
        environment:
            SERVER_NAME: ':80'
            MERCURE_PUBLISHER_JWT_KEY: '${MERCURE_PUBLISHER_JWT_KEY:-!ChangeThisDevMercureHubJWTSecretKey!}'
            MERCURE_SUBSCRIBER_JWT_KEY: '${MERCURE_SUBSCRIBER_JWT_KEY:-!ChangeThisDevMercureHubJWTSecretKey!}'
            # Set the URL of your Symfony project (without trailing slash!) as value of the cors_origins directive
            MERCURE_EXTRA_DIRECTIVES: |
                cors_origins http://app.domain.com
                use_forwarded_headers "1"
        # Comment the following line to disable the development mode
        command: /usr/bin/caddy run --config /etc/caddy/Caddyfile.dev
        volumes:
            - mercure_data:/data
            - mercure_config:/config

volumes:
    mercure_data:
    mercure_config:
```
Docker override:
```yaml
services:
###> symfony/mercure-bundle ###
  mercure:
    ports:
      - "8181:80"
###< symfony/mercure-bundle ###
```

```nginx
server {
    listen 443 ssl http2;
    listen [::]:443 ssl http2;

    server_name hub-preprod.domain.com;

    ssl_certificate path_to_crt;
    ssl_certificate_key path_to_key;

    location / {
        proxy_pass http://localhost:8181;
        proxy_read_timeout 24h;
        proxy_http_version 1.1;
        proxy_set_header Connection "";

        ## Be sure to set USE_FORWARDED_HEADERS=1 to allow the hub to use those headers ##
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Host $host;
        proxy_set_header X-Forwarded-Proto $scheme;
    }

    error_log /home/<user>/logs/error-hub.log;
[...]

PR #1005, opened 2025-02-06
docs: add proxy_buffering off to Nginx example for SSL proxy compatibility

erseco:
This PR updates the documentation to include `proxy_buffering off;` in the Nginx configuration example for Mercure. This change improves compatibility when using Mercure behind an SSL-enabled Nginx proxy.  

### Problem  
When Mercure is behind a Nginx reverse proxy with SSL, Nginx buffers responses by default. This can cause issues with long-lived SSE connections, leading to delayed message delivery and unexpected behavior for subscribers.  

### Solution  
Disabling Nginx buffering with the following directive ensures that SSE messages are forwarded immediately to clients:  

```nginx
proxy_buffering off;
```

### Changes  
- **Docs:** Updated the Nginx configuration example to include `proxy_buffering off;`  
- **Notes:** Added a brief explanation of why this directive is necessary when using SSL proxies.  

### Testing  
- Tested with Mercure behind an Nginx SSL proxy.  
- Observed that after adding `proxy_buffering off;`, SSE messages are delivered in real-time without delays.  

### Related issues
#854 #843 #857 #986

### DIFF
--- a/docs/hub/nginx.md
+++ b/docs/hub/nginx.md
@@ -16,6 +16,9 @@ server {
         proxy_http_version 1.1;
         proxy_set_header Connection "";
 
+        # Enable fast reply in SSE
+        proxy_buffering off;
+
         ## Be sure to set USE_FORWARDED_HEADERS=1 to allow the hub to use those headers ##
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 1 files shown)


dunglas (approved):

dunglas:
Thanks!

PR #1017, opened 2025-02-26
feat: Add token-level control for public updates 

MHNassar:
**Problem**
Currently, publishers can choose between private and public updates by setting the private parameter in their request. However, this presents a security risk in frontend applications where a malicious user could modify requests to remove the private flag, potentially exposing sensitive information to unauthorized subscribers.

**Solution**

This PR introduces a new JWT claim `allow_public_updates` that can be set to false to restrict tokens to private-only updates. When this claim is present and set to false, any attempt to publish a public update will be rejected with a 401 Unauthorized response, regardless of what's in the request.
This provides a server-enforced security boundary that cannot be bypassed by manipulating requests on the client side.

**Implementation**

Added a new `canDispatchPublic` function that checks for the presence of the claim
Integrated this check into the PublishHandler flow
Added tests to verify the functionality
Maintains backward compatibility by defaulting to allowing public updates when the claim is not present

**Use Case**

This feature is particularly useful for collaborative editing applications where you want to ensure that document updates are only visible to authorized collaborators, regardless of how the frontend code behaves.

### DIFF
--- a/authorization.go
+++ b/authorization.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/golang-jwt/jwt/v5"
 	"go.uber.org/zap"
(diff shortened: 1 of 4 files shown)


dunglas:
We need to patch the spec first. Maybe can we find something more idiomatic claim name?

Issue #1024, opened 2025-03-09
High memory usage with the latest version


divine:
Hello,

I've encountered an issue with high memory usage in the latest version. Mercure seems to hang and doesn't respond to publish commands. The average response time now takes up to 3 minutes when the memory is full, and it doesn’t seem to free up memory - it just keeps filling up without using SWAP.

Setting up `lru_cache 7e8 1e9` gives an error, I thought it was supported before?

Error: adapting config using caddyfile: parsing caddyfile tokens for 'route': parsing caddyfile tokens for 'mercure': strconv.ParseInt: parsing "7e8": invalid syntax, at /etc/caddy/Caddyfile:30

I don't know maybe it's because of the removal of lru cache #970. Also not sure why some of the bench test was removed?

Maybe we should debug it and try to fix the problem with high memory usage and make Mercure great again :)

Thanks!

dunglas:
I maybe made a mistake. I thought this cache was not used anymore.

Maybe should we revert this commit?

divine:
Hello,

I'm not against new caching libraries but probably we need to bench both of the libraries and see which one is performing better and give the user ability to choose which one they want to use. If they perform same on bench script then we might stick with new caching library.

I'll need to get memory dump to see why it's filling up. Maybe it's not a cache issue but some bug. Might be just a bench script could show the reason as well!

Currently there are more than 10k concurrent connections and the memory fills up pretty slowly in a day.

Thanks!

divine:
Hello,

Looks like we've got a memory leak! Still trying to figure it out. PR #971 probably when the leak was intoduced:

```sh
panic: close of closed channel

goroutine 3 [running]:
github.com/dunglas/mercure.(*LocalSubscriber).Disconnect(0x140001be140)
	/Users/divine/Documents/work/tests/mercure/localsubscriber.go:125 +0x90
github.com/dunglas/mercure.(*LocalTransport).Close.func1.1(0x18?)
	/Users/divine/Documents/work/tests/mercure/local.go:106 +0x1c
github.com/dunglas/mercure.(*LocalTransport).Close.func1.(*SubscriberList).Walk.2({0x102d7c2e0?, 0x140001be140?})
	/Users/divine/Documents/work/tests/mercure/subscriberlist.go:102 +0x40
github.com/kevburnsjr/skipfilter.(*SkipFilter).Walk(0x14000032200, 0x0, 0x14000183b50)
[...]

PR #1040, opened 2025-05-04
fix: LocalSubscriber hanging

dunglas:
Fixes https://github.com/dunglas/mercure/issues/1024


divine:
Hello,

I've now deployed Mercure with this PR but got the error:

```bash
Error: adapting config using caddyfile: parsing caddyfile tokens for 'route': parsing caddyfile tokens for 'mercure': strconv.ParseInt: parsing "1e5": invalid syntax, at /etc/caddy/Caddyfile:34
```

`lru_cache 1e5`

I was forced to remove it but I need it because of high amount of subscribers, hopefully this will be fixed as well!

Thanks!

dunglas:
Can you try `100000` instead?

dunglas:
I fixed the docs and added support for configuring both the number of shards and the number of items per shard.

localsubscriber.go
```diff
@@ -80,25 +70,16 @@ func (s *LocalSubscriber) Dispatch(u *Update, fromHistory bool) bool {
...
-
 	for _, u := range s.liveQueue {
 		select {
 		case s.out <- u:
 			n++
 		default:
-			s.handleFullChan()
-			s.liveMutex.Unlock()
-
 			return n
```
kevburnsjr, line 78: `handleFullChan` was calling `atomic.StoreInt32(&s.disconnected, 1)`. Removing this call changes the behavior. 

Could this not cause events to be silently unflushed when the channel is full, causing the client to enter a ready state after missing events?

localsubscriber.go
```diff
@@ -49,26 +46,19 @@ func (s *LocalSubscriber) Dispatch(u *Update, fromHistory bool) bool {
 	}
 
 	if !fromHistory && atomic.LoadInt32(&s.ready) < 1 {
-		s.liveMutex.Lock()
```
kevburnsjr, line 52: Is it safe to remove these mutexes?

localsubscriber.go
dunglas: In the meantime, I changed the code to use the context API instead, simplifying the code and fixing the race conditions showcased by the benchmark.

The existing behavior should be preserved, too.

localsubscriber.go
dunglas: Added back while removing contexts.

dunglas:
@divine I finally found the bug crashing the benchmark. It was in the benchmark itself.

Could you try this PR to see if you get improvements?

divine:
Hello,

I've deployed it now. I'll let you know if it fails.

Thank you guys for the quick fixes!

divine:
> @divine any news? I would like to merge this one!

Works fine for me! Didn't crash and  still working fine.

Thanks!

PR #1078, opened 2025-09-03
fix: update secret key ref to use existing secret

TaminoWalter:
The `CADDY_EXTRA_CONFIG` secret in the `deployment.yaml` of the helm chart references the wrong secret name when `existingSecret` is set in the `values.yaml`. 

Currently it is using `{{ include "mercure.fullname" . }}`, correct would be `{{ include "mercure.secretName" . }}`.

### DIFF
--- a/charts/mercure/templates/deployment.yaml
+++ b/charts/mercure/templates/deployment.yaml
@@ -50,7 +50,7 @@ spec:
             - name: CADDY_EXTRA_CONFIG
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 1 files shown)


dunglas (approved):

t-richard (approved):
We're facing the same issue and came up with the same fix

dunglas:
Thanks @TaminoWalter and @t-richard for the fix. I'm sorry for not getting back to you sooner.

PR #959, opened 2024-10-14
spec!: add support for several topic matcher types, including URL Pattern

dunglas:
Replaces https://github.com/dunglas/mercure/pull/831 with a more flexible and future-proof approach.

Closes #588.

## Summary

* In addition to URI templates, [regular expressions](https://datatracker.ietf.org/doc/rfc9485/), [URL patterns](https://urlpattern.spec.whatwg.org/) and [CEL](https://cel.dev/) expressions are now supported
* the `topic` query parameter becomes `match` and now only supports exact matching
* new query parameters `match{MatchType}` are introduced
  * by default, `matchRegexp`, `matchURLPattern`, `matchCEL`, and `matchURITemplate` are defined
  * hub can implement custom matcher types
  * future updates may introduce new matcher types
* it's now possible to include per-topic (or topic matcher) payloads (replaces #945)
* the JWT format has been updated to match these new capabilities
* the subscription API format has been updated to match these new capabilities

The official hub will be able to support both the current and new protocols simultaneously (no hard BC break).

TODO:

* [x] update all examples

PR #1207, opened 2026-04-18
feat!: topic matchers (Exact, URLPattern, Regexp, URITemplate, CEL)

dunglas:
Implements the topic-matcher revision of the Mercure spec in the
[`spec/topic-selectors`](https://github.com/dunglas/mercure/tree/spec/topic-selectors)
branch.

## Wire protocol

- Subscribers use case-insensitive `match*` query parameters: `match` / `matchExact`, `matchURLPattern`, `matchRegexp`, `matchURITemplate`, `matchCEL`.
- JWT `mercure.publish` / `mercure.subscribe` entries accept the legacy string form and the new object form `{match, matchType, payload}`.
- Unknown matcher types → 501 Not Implemented.
- Subscription API / events use `/.well-known/mercure/subscriptions/{matchType}/{match}/{subscriber}`. The v8 `{topic}[/{subscriber}]` routes are only registered under `WithProtocolVersionCompatibility`.
- Per-subscription payload is the first JWT subscribe-claim whose matcher accepts the subscription's pattern, falling back to `mercure.payload`.

## Go API

- One `Matcher` interface: `Match(topics []string, pattern string) bool`. Full topic list passed through so CEL's aggregate semantics (`topics.size()`, `topics.all()`, …) work.
- `WithMatcherType(name, impl)` is the only registration option. `NewHub` picks defaults based on compat mode: Exact + URLPattern (modern), Exact + URITemplate (legacy). Exact is always registered — the protocol mandates it.
- Built-in matchers: `ExactMatcher`, `URITemplateMatcher`, `URLPatternMatcher`, `RegexpMatcher` (I-Regexp, RFC 9485), `CELMatcher`.
- Internal `legacyMatcher` (exact + URI template) plugs the legacy `topic=` param and v8 string JWT claims into the same matcher pipeline.

## Topic selector store

- Single unsharded otter cache; struct key `(Type, Pattern, Topics)` where `Topics` is the update's list joined with NUL — zero-alloc for single-topic updates.
- Exact matching skips the cache (already O(1)).

## Subscriber

- Unified through `SubscribedMatchers` / `AllowedPrivateMatchers`. Deprecated v8 fields (`SubscribedTopics`, `AllowedPrivateTopics`, `EscapedTopics`, `legacyTopicParam`) and the `SetTopics` / `matchTopicsLegacy` / `getSubscriptionsLegacy` helpers are removed. Legacy flows are now just `SetMatchers` with `legacyMatcher`-backed `TopicMatcher`s.

## Security

- CEL runs with a cost limit; over-budget evaluations return false instead of burning CPU on pathological expressions from hostile JWTs.
- Matcher pattern length and per-request matcher count are bounded.
- `RegisterMatcherType` documented as setup-time-only.

Refs: #959


dunglas:
Heads-up: the spec this implements has changed substantially in https://github.com/dunglas/mercure/pull/1262. The matcher set is reduced to two types — `Exact` and `URLPattern` — so `Regexp`, `URITemplate`, and `CEL` (and the matcher-type registry) are removed. Authorization also moves from the `mercure` claim to OAuth 2.0 `authorization_details` (RFC 9396), and alternate topics are dropped (one topic per update). This PR needs rework against the new spec before it can land. Leaving it open as a tracking reference for now.

dunglas:
Superseded by https://github.com/dunglas/mercure/pull/1269, which implements the reduced matcher surface of https://github.com/dunglas/mercure/pull/1262 (Exact + URLPattern only, no matcher registry) with the v8 behaviors behind the `deprecated_topic` build tag.

PR #1269, opened 2026-06-07
feat!: Exact and URLPattern topic matchers

dunglas:
Implements the topic-matcher portion of the revised specification (https://github.com/dunglas/mercure/pull/1262) in the hub. Supersedes https://github.com/dunglas/mercure/pull/1207.

## Wire protocol

- Subscribers use the case-sensitive `match` (exact matching, the default; `match_exact` is the explicit spelling) and `match_urlpattern` ([WHATWG URL Pattern](https://urlpattern.spec.whatwg.org/)) query parameters, mirroring the optional `match_type` member of the token's topic matchers. Any other name under the reserved `match` prefix (compared ASCII case-insensitively, so case typos fail loudly) is rejected with a 400, as are values containing control characters (C0, C1, DEL), invalid UTF-8, or exceeding the per-request limits.
- URL Patterns are resolved against the hub URL (new `public_url` Caddyfile directive / `WithPublicURL` option); `ignoreCase` is never enabled, and compiled patterns are memoised.
- An update has exactly one topic: publish requests with several `topic` fields get a 400. Topics containing control characters or invalid UTF-8 are rejected, and the reserved `/.well-known/mercure` namespace test follows the spec: the topic is resolved as a URL reference against the hub URL (WHATWG algorithm, the same canonicalization as URL Pattern matching) and percent-encoding of unreserved characters is normalized before the path check, so relative topics, backslashes, and `%6D`-style encodings cannot forge subscription events. Publisher-supplied `id` values starting with `#` or equal to the reserved `earliest` are rejected.
- JWT `mercure.publish` / `mercure.subscribe` entries use the object form `{"match": "...", "match_type": "exact|urlpattern", "payload": ...}`; `match_type` is case-sensitive and defaults to `exact`. The per-subscription payload is the first matching subscribe claim's payload, falling back to `mercure.payload`.
- The subscription API moves to `/.well-known/mercure/subscriptions/{match_type}/{match}/{subscriber}`, documents expose `match` / `match_type` instead of `topic`, and `ETag` values are RFC 9110 quoted strings.

## Differences with #1207

Per the matcher reduction in #1262, the exploratory surface is gone:

- No `Matcher` interface, no matcher registry, no `WithMatcherType`, no `matcher_types` directive. Matcher types are a closed string enum (`MatcherType`); matching and validation are plain `switch`es.
- Regexp, URI Template and CEL matchers are dropped (no cel-go/antlr dependencies). URI Template matching survives only inside the deprecated v8 matcher.
- The subscribe parameters follow the final spec naming: `match` / `match<MatcherType>`.

## Deprecation layer

The v8 behaviors follow the existing `deprecated_server` / `deprecated_transport` conventions, behind a new `deprecated_topic` build tag and gated at runtime by `WithProtocolVersionCompatibility(8)` / `protocol_version_compatibility 8`:

- URI Template selectors in the `topic` query parameter,
- bare-string JWT matcher claims,
- alternate topics on updates (`Update.Topics`, kept as a tag-gated embedded field next to the new `Update.Topic`),
- the `/subscriptions/{topic}[/{subscriber}]` routes.

Official binaries and Docker images keep the tag enabled; CI tests both build modes. `Update`'s JSON codec preserves the historic `"Topics"` array shape so bolt databases written by 0.x hubs stay readable.

## Breaking changes

- v8 string-array JWT claims and URI Template selectors are rejected (401/400) unless the hub is built with `deprecated_topic` and runs in compatibility mode; the hub logs a migration hint.
- Go API: `Update.Topics` becomes `Update.Topic`; `Subscriber.SetTopics` and the topic-selector fields are replaced by matcher-based equivalents.

See `docs/UPGRADE.md` for the migration guide.

## Performance

The two-matcher model was designed to be free on the hot path. Benchmarks on an Apple M5 Pro.

**Match path** (`BenchmarkMatchMatcher`, per matcher evaluation):

| matcher | time | allocs |
|---|---|---|
| exact, hit | 2.3 ns | 0 |
| exact, miss | 2.9 ns | 0 |
| wildcard (`*`) | 1.1 ns | 0 |
| URL Pattern, cached | 52 ns | 0 |
| URL Pattern, cold compile | 22 µs | (once per pattern, subscribe-time only) |

**No regression vs the previous implementation.** End-to-end subscriber dispatch (`BenchmarkSubscriber`, main vs this branch) is statistically unchanged: 136.4 ns → 136.5 ns/op, 1 alloc/op (`p=0.78`).

**URL Pattern compilation is now ~21% faster** (‑25% memory, ‑22% allocs) via dunglas/go-urlpattern#12, cutting subscribe latency and cache-miss cost for `urlpattern` matchers.

**Load test** (Gatling, Caddy hub): zero failed requests across scenarios; end-to-end mean 32 ms, and a heavier run sustained 6,625 requests with 0 errors. A CPU profile under load shows the matcher path accounting for **under 1%** of hub CPU — the hub is network-I/O and fsync bound, not match bound.

### How fast is that?

- **Exact topic matching runs in ~2 ns with zero allocations** — a single core evaluates hundreds of millions of topic matches per second.
- **URL Pattern matches resolve from cache in ~50 ns** — roughly 19 million matches per second per core — after a one-time compile.
- **Matching never shows up in CPU profiles under load.** Mercure's throughput scales with your network and your hardware, not with how many topics or how complex your matchers are. Add cores, add bandwidth — the matcher gets out of the way.

Refs https://github.com/dunglas/mercure/pull/1262
Refs https://github.com/dunglas/mercure/issues/959


dunglas:
Consolidated into #1278, which now carries the matchers, OAuth 2.0, and docs in a single PR to main.

PR #1273, opened 2026-06-08
feat!: OAuth 2.0 authorization (RFC 9068/9396/6750/9728)

dunglas:
Implements the authorization half of the revised specification (https://github.com/dunglas/mercure/pull/1262), stacked on the topic-matcher PR https://github.com/dunglas/mercure/pull/1269 (base `feat/matchers`). Retarget to `main` once #1269 merges.

## Wire protocol

- The hub is an OAuth 2.0 protected resource. Access tokens are validated as JWT access tokens ([RFC 9068](https://www.rfc-editor.org/rfc/rfc9068)): `typ` header `at+jwt`, `aud` containing the hub's resource identifier, `exp` required. Configure the audience with `resource_identifier` (Caddyfile) / `WithResourceIdentifier` (Go); it defaults to the public URL.
- Authorization is expressed with the standard `authorization_details` claim ([RFC 9396](https://www.rfc-editor.org/rfc/rfc9396)): `{"type":"mercure","actions":["publish"|"subscribe"],"topics":[{"match","matchType?"}],"payload?":{}}`. One invalid `mercure` detail rejects the whole token as `invalid_token` (401, no partial acceptance — and no disclosure that the signature verified but the claims were malformed); the number of details, topics per detail and pattern length are capped.
- Errors follow [RFC 6750](https://www.rfc-editor.org/rfc/rfc6750): 401 + bare `WWW-Authenticate: Bearer` challenge (with a `resource_metadata` parameter) when no token is presented, 401 `invalid_token` on validation failure, 403 `insufficient_scope` when a valid token lacks the action on the topic, 400 `invalid_request` for malformed requests.
- The access token is presented via `Authorization: Bearer` (scheme matched case-insensitively per RFC 9110), the `access_token` query parameter (renamed from `authorization`), or the cookie (renamed `mercureAccessToken`, configurable). Precedence: header, then query parameter, then cookie.
- OAuth 2.0 Protected Resource Metadata ([RFC 9728](https://www.rfc-editor.org/rfc/rfc9728)) is served at `/.well-known/oauth-protected-resource/.well-known/mercure`; `authorization_servers` advertises issuers and the `mercure_cookie` member advertises cookie support.
- Unauthenticated subscribers remain supported (`anonymous`): they receive only non-private updates, per the spec. CORS credentials are disabled when `cors_origins` contains `*`, since the protocol forbids combining a wildcard `Access-Control-Allow-Origin` with credentials.

## Deprecation layer (`deprecated_claim`)

The legacy `mercure` claim (string and object forms), the `https://mercure.rocks/` namespaced claim, `mercure.payload`, the `authorization` query parameter and tokens without `typ: at+jwt`/`aud` move behind the new `deprecated_claim` build tag, honored only with `protocol_version_compatibility 8`. Official binaries and Docker images keep the tag; CI tests both build modes.

## Breaking changes

- `mercure`-claim tokens are rejected (401) unless the hub is built with `deprecated_claim` and runs in compatibility mode.
- The official Caddyfile now sets `resource_identifier` (default `https://localhost/.well-known/mercure`); set `MERCURE_RESOURCE_IDENTIFIER` to the audience your tokens carry, or enable `protocol_version_compatibility 8`.
- Go API: `canReceive`/`canDispatch` are replaced by the internal authorization-detail grant logic; `NewHub` requires a resource identifier (or public URL) when JWT auth is enabled in modern mode.

See `docs/UPGRADE.md`.

RFC 8414 authorization-server key discovery is intentionally left as a follow-up (the existing `publisher_jwks_url`/`subscriber_jwks_url` directives already cover external key distribution).

Refs https://github.com/dunglas/mercure/pull/1262


dunglas:
Consolidated into #1278, which now carries the matchers, OAuth 2.0, and docs in a single PR to main.

PR #1294, opened 2026-07-06
Enable HTTPS support for Mercure

jza0330:
## Summary

This PR introduces several improvements to the Helm chart that make it possible to expose Mercure over HTTP/2.  Running HTTPS in the Mercure pod is required to enable HTTP/2 between clients and the application, allowing HTTP/2 traffic to be preserved end-to-end.

## Changes

* Move non-sensitive runtime configuration from the `Secret` to a `ConfigMap`, keeping only sensitive values in the `Secret`.
* Add support for additional `volumes` and `volumeMounts` in the Deployment template, making it possible to mount TLS certificates or other runtime resources.
* Make the main Service port name configurable via Helm values, so you can name it `https` instead of `http`
* Allow separate annotations for the main Service and the metrics Service, enabling independent configuration (for example, different OpenShift or cloud-provider annotations).

## Motivation

When deploying Mercure on OpenShift, using a passthrough route allows TLS termination to occur in the Mercure pod instead of at the router. This preserves end-to-end HTTPS and enables HTTP/2 communication between clients and the application, which is important for Mercure's real-time capabilities.

The changes in this PR provide the flexibility needed to configure such deployments while also improving the overall configurability and maintainability of the Helm chart.


dunglas:
Really nice work on this, thank you! The HTTP/2 passthrough use case is a great addition, and the extra volumes, configurable `portName`, and splitting the non-sensitive Caddy directives into a ConfigMap are all solid improvements. A few things I'd like to sort out before merging:

**1. Let's keep `transport-url` in the Secret.** This one's a security thing rather than a style preference: Redis and PostgreSQL transport URLs usually carry inline credentials (`redis://default:PASSWORD@host`, `postgres://user:PASS@host`), so moving the key to the ConfigMap would expose the password to anyone with `get configmap` on the namespace — the same leak we document for the Redis password in #1313. Could you leave `transport-url` in `secrets.yaml` with its `secretKeyRef` wiring? The `{env.NAME}` placeholder trick sadly isn't a reliable substitute here, since `transport_url` is parsed by the Mercure module and isn't guaranteed to run through Caddy's placeholder replacer the way the Caddyfile-text directives do. The other three keys (`extra-directives`, `caddy-extra-config`, `caddy-extra-directives`) are perfectly fine to move.

**2. Could we keep the metrics Service annotations working?** The `service.yaml` change drops the `service.annotations` block from the metrics Service, so anyone annotating both would lose that. Either keep applying `service.annotations` to both, or add a dedicated `metricsService.annotations` value — whichever feels cleaner to you.

**3. A small docs touch-up.** The README table already picks up the new values (nice), but it'd be great to mention the HTTPS/HTTP/2 passthrough workflow in `docs/deployment/kubernetes.md` — the TLS cert volumes and `service.portName: https` — plus a line on the ConfigMap vs Secret split so folks know credentials belong in the Secret.

Thanks again for pushing this forward!

jza0330:
> Really nice work on this, thank you! The HTTP/2 passthrough use case is a great addition, and the extra volumes, configurable `portName`, and splitting the non-sensitive Caddy directives into a ConfigMap are all solid improvements. A few things I'd like to sort out before merging:
> 
> **1. Let's keep `transport-url` in the Secret.** This one's a security thing rather than a style preference: Redis and PostgreSQL transport URLs usually carry inline credentials (`redis://default:PASSWORD@host`, `postgres://user:PASS@host`), so moving the key to the ConfigMap would expose the password to anyone with `get configmap` on the namespace — the same leak we document for the Redis password in #1313. Could you leave `transport-url` in `secrets.yaml` with its `secretKeyRef` wiring? The `{env.NAME}` placeholder trick sadly isn't a reliable substitute here, since `transport_url` is parsed by the Mercure module and isn't guaranteed to run through Caddy's placeholder replacer the way the Caddyfile-text directives do. The other three keys (`extra-directives`, `caddy-extra-config`, `caddy-extra-directives`) are perfectly fine to move.
> 
> **2. Could we keep the metrics Service annotations working?** The `service.yaml` change drops the `service.annotations` block from the metrics Service, so anyone annotating both would lose that. Either keep applying `service.annotations` to both, or add a dedicated `metricsService.annotations` value — whichever feels cleaner to you.
> 
> **3. A small docs touch-up.** The README table already picks up the new values (nice), but it'd be great to mention the HTTPS/HTTP/2 passthrough workflow in `docs/deployment/kubernetes.md` — the TLS cert volumes and `service.portName: https` — plus a line on the ConfigMap vs Secret split so folks know credentials belong in the Secret.
> 
> Thanks again for pushing this forward!

your comments incorporated into the code

dunglas:
Thanks for your work, we're almost there!

`templates/configmap.yaml` renders `extra-directives`, `caddy-extra-config`, `caddy-extra-directives` unconditionally from `.Values.*`, while `templates/deployment.yaml` now sources those three env vars via `configMapKeyRef` instead of `secretKeyRef` against `mercure.secretName`.

Consequence: anyone using `existingSecret` to supply a custom secret containing `extra-directives` / `caddy-extra-directives` (as `values.yaml`'s own comment block still documents) will have those values silently ignored — the chart now always uses its own rendered ConfigMap value (default `""`) instead. Functional regression, not just a docs nit.

The comment block under `existingSecret` in `values.yaml` (~line 38-44) still lists `extra-directives:` and `caddy-extra-directives:` as required secret keys — no longer true, should be updated. Worth confirming whether this is an intentional breaking change.

jza0330:
> Thanks for your work, we're almost there!
> 
> `templates/configmap.yaml` renders `extra-directives`, `caddy-extra-config`, `caddy-extra-directives` unconditionally from `.Values.*`, while `templates/deployment.yaml` now sources those three env vars via `configMapKeyRef` instead of `secretKeyRef` against `mercure.secretName`.
> 
> Consequence: anyone using `existingSecret` to supply a custom secret containing `extra-directives` / `caddy-extra-directives` (as `values.yaml`'s own comment block still documents) will have those values silently ignored — the chart now always uses its own rendered ConfigMap value (default `""`) instead. Functional regression, not just a docs nit.
> 
> The comment block under `existingSecret` in `values.yaml` (~line 38-44) still lists `extra-directives:` and `caddy-extra-directives:` as required secret keys — no longer true, should be updated. Worth confirming whether this is an intentional breaking change.

yes, this is an intentional breaking change otherwise the code wouldn't be nice and maintainable.
Where do you want me to document this?  `UPGRADE.md` ?

jza0330:
@dunglas Is there anything else I could do or improve to help us move this PR forward?

PR #1299, opened 2026-07-18
docs: complete RFC 9068 token example

J3m5:
## Summary

Add the four RFC 9068 claims still missing from the main JWT access token example in `docs/concepts/authorization.md`:

- `sub`
- `client_id`
- `iat`
- `jti`

The surrounding text already states that RFC 9068 requires issuers to populate these claims, and the specification example was completed in #1297. This keeps the user-facing documentation example consistent with both RFC 9068 §2.2 and `spec/mercure.md`.

## Scope

Documentation only. No runtime behavior changes.

### DIFF
--- a/docs/concepts/authorization.md
+++ b/docs/concepts/authorization.md
@@ -22,8 +22,12 @@ A Mercure access token is a JWT access token as defined by [RFC 9068](https://ww
 // payload
 {
   "iss": "https://example.com",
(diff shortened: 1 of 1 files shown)


dunglas (approved):

dunglas:
Thank you very much for reviews and patches!!

PR #1365, opened 2026-08-24
feat: add drain_timeout for graceful shutdown draining

pseidemann:
Decouple the graceful shutdown drain window from write_timeout: a long write_timeout keeps steady-state reconnects rare while a shorter drain_timeout lets rolling updates finish quickly. On termination each subscriber's clean close is rescheduled to a random point spread uniformly across the window; config reloads never drain (gated on caddy.Exiting()), keeping reloads reconnect-free. Works with write_timeout 0 too.

Wires the drain_timeout Caddyfile directive, updates the docs, and adjusts the Helm chart grace-period guidance to key on the drain window.


subscribe.go
```diff
@@ -148,6 +149,16 @@ func (h *Hub) SubscribeHandler(w http.ResponseWriter, r *http.Request) {
 		heartbeatTimerC = heartbeatTimer.C
 	}
 
+	// disconnectionTimer is created lazily and stopped through a nil-checking
+	// defer: a subscriber with neither a write timeout nor a token exp starts
+	// with no deadline (disconnectionTimerC stays nil), yet the drain path may
+	// still arm it on shutdown, so its creation can't be pinned to a defer here.
+	defer func() {
+		if disconnectionTimer != nil {
```
dunglas: Isn't this creating a potential race condition?

dunglas:
The idea looks sensible to me. I'll do a careful review early September when I'll be back from vacation.

dunglas:
I found three issues in the new drain path:

1. **Apply the shortened deadline to the socket.** Updating `rc.writeDeadline` only changes the stored value. With `dispatch_timeout 0`, a subsequent blocked write retains the original socket deadline and prevents the handler from processing its disconnection timer. A focused reproduction with a three-minute drain remained blocked more than 16 minutes beyond that window. The drain branch should apply the shortened deadline through the response controller. [subscribe.go:229–239](https://github.com/dunglas/mercure/blob/0a34105bc477b20d5d167477239b4b1b2cb7db56/subscribe.go#L229-L239)

2. **Preserve an earlier JWT write deadline.** When expiry is closer than `dispatch_timeout`, the existing disconnection time equals the write deadline. Replacing the socket deadline with `deadline.Add(h.dispatchTimeout)` can then extend it beyond expiry. With a token expiring in two seconds and the default five-second dispatch timeout, I reproduced a socket deadline three seconds beyond expiry. Clamp the replacement to the existing nonzero write deadline. [subscribe.go:229–233](https://github.com/dunglas/mercure/blob/0a34105bc477b20d5d167477239b4b1b2cb7db56/subscribe.go#L229-L233)

3. **Reject negative drain timeouts.** `WithDrainTimeout(-time.Second)` succeeds, but draining subsequently panics in `rand.Int64N`. Negative values should fail configuration rather than panic in subscriber handlers during shutdown. [hub.go:168–172](https://github.com/dunglas/mercure/blob/0a34105bc477b20d5d167477239b4b1b2cb7db56/hub.go#L168-L172)

The existing tests for both modules, focused shutdown race tests, and gopls checks pass; separate reproductions exposed these cases.

PR #1366, opened 2026-08-26
ci: fix golangci-lint v2.13 findings

pseidemann:
golangci-lint latest (v2.13.1) deprecated exhaustruct in favor of exhaustruct_v5; default:all enables both, so the disable list must name both. Also apply the new modernize errorsastype suggestion (errors.As -> errors.AsType) and a gofumpt blank-line fix in the caddy module.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,6 +32,7 @@ linters:
     - depguard
     - errcheck
     - exhaustruct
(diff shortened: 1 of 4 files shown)


dunglas (approved):

dunglas:
Thank you!

PR #1375, opened 2026-09-07
docs: eight corrections to the 1.0 documentation

BySplashGm:
Eight corrections to the 1.0 documentation, found while migrating
[`api-platform/esa`](https://github.com/api-platform/esa) onto `v1.0.0-alpha.3` —
hub, Caddyfile, tokens, JS client and a Symfony/API Platform app. Seven commits, one
per finding except the two custom-build ones, which touch the same paragraphs. Each is
independent, so any one can be dropped without touching the rest.

Worth saying first: the 1.0 docs are good. `UPGRADE` is a real migration guide with a
find-and-replace checklist, and every breaking change we had reverse-engineered from
the Go and PHP sources was already documented there. What follows is the residue.

### 1. `docker run dunglas/mercure` gives you a 0.x hub

`latest` tracks the newest stable release and GoReleaser suppresses rolling tags on
prereleases, so `latest` and `v0.24.2` share a digest today (`sha256:916834e4…`) while
`v1.0.0-alpha.3` is another (`sha256:16485a0b…`). The quickstart says
`docker run dunglas/mercure`, then subscribes with `match=`, publishes with an `at+jwt`
token and points at `caddy mercure-token --dev`. All three fail against the hub the
page just started, with nothing indicating the version is the problem.

Pins `v1.0.0-alpha.3` on the pages that then exercise 1.0 (quickstart, installation,
GitHub Actions) and says why. Deployment pages, where the version is the reader's
choice, keep the untagged image; `deployment/docker`'s image-variants list states the
rolling-tag policy instead.

**The cost, stated plainly:** 12 literal `v1.0.0-alpha.3` strings in `docs/` — seven
image tags, three Go module versions, two in prose — and alpha.1 through alpha.3
shipped inside five days. Mitigated, not free: every prose mention says "until 1.0 is
stable" and `grep -rn 'v1.0.0-alpha.3' docs/` finds all twelve, so a bump is one `sed`.
Nothing cheaper exists today — Docker Hub carries no floating 1.0 tag (only the exact
`v1.0.0-alpha.N` and their per-arch variants; `.goreleaser.yml` suppresses the rolling
tags on prereleases on purpose) and Go rejects a prerelease prefix query
(`.../caddy@v1.0.0-alpha` is an "unknown revision", `@latest` resolves to `v0.24.2`).
Which puts the clean fix in the release config rather than this PR: one floating
prerelease tag, `:next` or `:v1.0.0-alpha`, is a line of `.goreleaser.yml` and the docs
could then name it once and never churn. If that is preferred, this commit can shrink
to its prose half, or drop entirely — it is first on the branch and nothing else needs
it.

### 2. `xcaddy` builds a 0.x hub, and the Go floor is unstated

`installation` shows `xcaddy build --with github.com/dunglas/mercure/caddy`. Unpinned,
Go resolves the module to its newest stable tag, which the module proxy reports as
`v0.24.2` (2026-06-01) — so the documented command silently builds a 0.x hub.
`caddy/go.mod` at `v1.0.0-alpha.3` also declares `go 1.26`, which is unstated;
`caddy:2.8-builder` ships Go 1.23 and fails outright.

### 3. No recipe for a compatibility-mode build

`UPGRADE` says "Custom builds must pass the tags to `go build`" but never shows how,
and nothing connects it to `xcaddy` — the build path `installation` recommends, whose
escape hatch is the non-obvious `XCADDY_GO_BUILD_FLAGS`. Adds the one-liner to both
pages, and notes that the tags only make the behaviors available: the Caddyfile still
needs `protocol_version_compatibility`.

### 4. `protocol_version_compatibility 7` was not removed in 1.0

The 0.14 upgrade notes state the directive "has since been removed in 1.0". It has not:

- `caddy/mercure.go` rejects the value only when `v != 7 && v != 8`, with the error
  "compatibility mode only supports protocol versions 7 and 8";
- `WithProtocolVersionCompatibility` accepts `case 7, 8`;
- `isBackwardCompatiblyEnabledWith` tests `version >= o.protocolVersionCompatibility`,
  so `7` restores every `8` behavior as well as its own.

Two behaviors are reachable only with `7`: the `Last-Event-ID` query parameter
(`subscribe.go`) and publishing a public update to a topic the token does not grant
`publish` on (`publish.go`). Neither lives in a `_deprecated.go` file, so both compile
into every build and cannot be excluded the way `deprecated_topic` and
`deprecated_claim` can — setting `7` to reach one of them relaxes more than an operator
may expect.

Corrects the claim and documents what `7` relaxes, while steering readers to `8`. If
the docs were stating the intended end state, the opposite fix applies — remove `7`
from the code — and this commit should be dropped in favor of that.

### 5. How a subscriber attributes an update to a topic is shown, never stated

Multiplexing many topics over one connection is the headline of 1.0, and "which topic
is this update for?" is the first question it raises. `concepts/subscribing` says twice
that one connection carries any number of subscriptions and documents the frame as
`id` / `event` / `data` — no topic field — then never closes the loop. The answer
appears only as an example on `use-cases/live-data`, where the discriminator happens to
sit in the payload; the FAQ does not cover it either.

States it next to the frame it concerns: the payload is the discriminator, JSON-LD
`@id` is the natural field, a payload carrying no identifier cannot be placed, and the
publish `type` is available for coarse routing. Also notes why the hub cannot supply
the topic itself — matchers OR together, and alternate topics make the attribution
ambiguous by design.

### 6. `rel="self"` is missing from `concepts/discovery`

The discovery page covers only `rel="mercure"`. It opens by saying a client needs two
things — the hub URL and the authorization requirements — and never mentions the third:
which topic to subscribe to. The spec devotes a section to it (Topic Discovery): "The
publisher **MAY** include one Link Header with `rel=self` (the self link header). It
**SHOULD** contain the canonical URL for the topic. If the link with `rel=self` is
omitted, the current URL of the resource **MAY** be used as a fallback", followed by
content-negotiation rules built entirely on that link. Documenting only `rel="mercure"`
reads as though 1.0 dropped it.

`concepts/subscribing` had the same gap from the other side: its discovery example
subscribed to `Content-Location ?? res.url`, which the spec's discovery section does
not mention. Both pages now read the self link with the spec's fallback.

### 7. An embedded hub is pinned to its host's Mercure version

`concepts/publishing` and the FAQ both point at FrankenPHP as *the* example of a
framework shipping its own hub, and neither says which protocol version that hub
speaks. FrankenPHP's `go.mod` on `main` pins `github.com/dunglas/mercure v0.24.2` as of
v1.12.7, so its embedded hub is 0.x — and that is the default way to run a Symfony or
API Platform app with an embedded hub, since symfony-docker and the API Platform
distribution both build on FrankenPHP. A reader who follows these pages onto 1.0 hits a
hub that rejects 1.0 matchers and tokens, and no amount of upgrading on the PHP side
changes it. Says so, and points at the two ways out.

### 8. The 1.0 section never says the resumption parameter was renamed

`lastEventID` became `last_event_id` in 1.0, and the 1.0 half of this page does not
mention it: not in "What changed at a glance", not in "Migrate your subscribers", and
not in the find-and-replace checklist, whose only `Last-Event-ID` entry is about the
**response** header. `spec/mercure.md`'s own migration table has the row —
`lastEventID` -> `last_event_id` — so the two documents disagree.

The one mention in this file sits under the historical **0.14** heading and names the
wrong source parameter: "The `Last-Event-ID` query parameter was renamed
`last_event_id`". 0.14 renamed it to `lastEventID` — "replace all occurrences of the
`Last-Event-ID` query parameter with `lastEventID`" (`0.x:docs/UPGRADE.md`), and the
0.x hub reads `query.Get("lastEventID")` (`0.x:subscribe.go`). `last_event_id` is the
1.0 name (`subscribe.go`, `retrieveLastEventID`). Under 1.0 there are four names and
four fates:

| Sent by the client | Under 1.0 |
| --- | --- |
| `last_event_id` query parameter | works |
| `Last-Event-ID` request header | works, name unchanged |
| `Last-Event-ID` query parameter | `protocol_version_compatibility 7` only, and logged either way |
| `lastEventID` query parameter | never read, in any mode, and never logged |

So the name every up-to-date 0.x client uses — the one 0.x's own upgrade note and
deprecation message told it to adopt — is the single one with no compatibility path and
no log line, while the older, already-deprecated name has both. A client
find-and-replaces on this page, finds nothing about `lastEventID`, ships, and stops
replaying silently. `symfony/mercure`'s Twig helper is exactly that client: it emits
`&lastEventID=` regardless of the configured protocol version, and its comment cites
this very 0.14 note. Filed separately.

Adds the table row, a subscriber bullet and a checklist entry; corrects the 0.14 note;
and corrects finding 4's own bullet, which had inherited the same mistake — the behavior
`7` restores is the *pre*-0.14 `Last-Event-ID` parameter, not `lastEventID`.

### Not in this PR

- **No Symfony or API Platform page in the 1.0 tree.** The biggest gap we found, but a
  page to write rather than a correction, so it is a separate PR.
- **`/spec` is labelled "v0.x (stable)" while serving the 1.0 text**, and
  `reference/protocol` sends every normative question to it, so a reader who trusts the
  selector either reads 1.0 semantics as 0.x or dismisses the page. The label does not
  come from this repository — there is no site or nav configuration here, and
  `spec/mercure.md`'s own `draft-dunglas-mercure-08` is correct (`0deb240` bumped it
  from 07 in the 1.0 spec pass) — so it is a mercure.rocks fix, flagged not patched.

### Verification

Every finding was checked against `main` at `abcdfb7`, not the rendered site, and the
code claims cite files in this repository. Re-checked live: the Docker Hub tag list and
the manifest digests for `latest`, `v0.24.2` and `v1.0.0-alpha.3`; `go list -m` for
`github.com/dunglas/mercure/caddy` at `@latest` and at the `@v1.0.0-alpha` prefix;
FrankenPHP's `go.mod` at `v1.12.7` and on `main`. Finding 8 was checked on both sides of
the rename — `0.x:docs/UPGRADE.md` and `0.x:subscribe.go` for the 0.14 name, `main`'s
`retrieveLastEventID` for the 1.0 one, and a grep of `main` confirming no code path
reads `lastEventID`.

Docs only — no code, no behavior change. Relative links and heading anchors across
`docs/` were checked to resolve. `prettier`, `markdownlint` and `textlint` have not been
run; their output will be applied on request.

### DIFF
--- a/docs/concepts/publishing.md
+++ b/docs/concepts/publishing.md
@@ -185,6 +185,8 @@ For stricter delivery guarantees (every state change reaches the hub even if the
 
 The Mercure protocol does not require an external hub. An application that already terminates HTTP/2 connections can speak the protocol directly: write SSE bytes to subscribers, validate JWTs, run matchers. This is unusual outside of frameworks that ship their own hub (FrankenPHP, for instance), but the spec allows it.
 
+Check which version of the protocol an embedded hub speaks before you rely on it. [FrankenPHP](https://frankenphp.dev) 1.12.7 embeds `github.com/dunglas/mercure v0.24.2`, so an application running its built-in hub — which includes anything built on the symfony-docker or API Platform distributions — is on 0.x and cannot opt into 1.0 by upgrading the PHP side. Reaching 1.0 there means either running a standalone hub next to the app, or building a custom binary with the 1.0 Caddy module (see [Custom Caddy build](../getting-started/installation.md#custom-caddy-build)).
+
 For everyone else, run the hub.
 
 ## Next steps for Mercure publishing
--- a/docs/concepts/publishing.md
+++ b/docs/concepts/publishing.md
@@ -185,8 +185,6 @@ For stricter delivery guarantees (every state change reaches the hub even if the
 
 The Mercure protocol does not require an external hub. An application that already terminates HTTP/2 connections can speak the protocol directly: write SSE bytes to subscribers, validate JWTs, run matchers. This is unusual outside of frameworks that ship their own hub (FrankenPHP, for instance), but the spec allows it.
 
-Check which version of the protocol an embedded hub speaks before you rely on it. [FrankenPHP](https://frankenphp.dev) 1.12.7 embeds `github.com/dunglas/mercure v0.24.2`, so an application running its built-in hub — which includes anything built on the symfony-docker or API Platform distributions — is on 0.x and cannot opt into 1.0 by upgrading the PHP side. Reaching 1.0 there means either running a standalone hub next to the app, or building a custom binary with the 1.0 Caddy module (see [Custom Caddy build](../getting-started/installation.md#custom-caddy-build)).
-
 For everyone else, run the hub.
 
 ## Next steps for Mercure publishing
--- a/docs/concepts/subscribing.md
+++ b/docs/concepts/subscribing.md
@@ -184,7 +184,7 @@ es.onmessage = (event) => {
 };
 ```
 
-JSON-LD's `@id` is the natural field because it is the topic URL itself, but any identifier works as long as every publisher of that topic sets it. Publishing a bare value with no identifier in it — a number, a status string — leaves a multi-topic subscriber unable to place the update.
+JSON-LD's [`@id`](https://www.w3.org/TR/json-ld11/#node-identifiers) is the natural field because it is the topic URL itself, but any identifier works as long as every publisher of that topic sets it. Publishing a bare value with no identifier in it — a number, a status string — leaves a multi-topic subscriber unable to place the update.
 
 For coarse routing, the publish `type` becomes the SSE `event` field, so a family of updates can be picked up with `addEventListener("<type>", ...)` instead. That is a label you choose, not the topic, and `mercure` is reserved for [subscription events](active-subscriptions.md).
 
@@ -202,7 +202,6 @@ Host: example.com
 200 OK
 Content-Type: application/ld+json
 Link: <https://hub.example.com/.well-known/mercure>; rel="mercure"
-Link: </books/1>; rel="self"
 ```
 
 Subscribers that fetch the resource first can read the headers to find both the hub and the topic to subscribe to:
@@ -213,10 +212,9 @@ const res = await fetch("https://example.com/books/1");
 const links = res.headers.get("Link");
 
 const hub = links.match(/<([^>]+)>;\s*rel="?mercure"?/)[1];
-const self = links.match(/<([^>]+)>;\s*rel="?self"?/)?.[1] ?? res.url;
 
 const url = new URL(hub);
-url.searchParams.append("match", new URL(self, res.url).toString());
+url.searchParams.append("match", res.url);
 new EventSource(url);
 ```
 
--- a/docs/concepts/subscribing.md
+++ b/docs/concepts/subscribing.md
@@ -202,21 +202,21 @@ Host: example.com
 200 OK
 Content-Type: application/ld+json
 Link: <https://hub.example.com/.well-known/mercure>; rel="mercure"
+Link: </books/1>; rel="self"
 ```
 
-Subscribers that fetch the resource first can read the header to find the hub:
+Subscribers that fetch the resource first can read the headers to find both the hub and the topic to subscribe to:
 
 ```javascript
 // Discovering the Mercure Hub via Link Header
 const res = await fetch("https://example.com/books/1");
-const link = res.headers.get("Link");
-const hub = link.match(/<([^>]+)>;\s*rel="?mercure"?/)[1];
+const links = res.headers.get("Link");
+
+const hub = links.match(/<([^>]+)>;\s*rel="?mercure"?/)[1];
+const self = links.match(/<([^>]+)>;\s*rel="?self"?/)?.[1] ?? res.url;
 
 const url = new URL(hub);
-url.searchParams.append(
-  "match",
-  res.headers.get("Content-Location") ?? res.url,
-);
+url.searchParams.append("match", new URL(self, res.url).toString());
 new EventSource(url);
 ```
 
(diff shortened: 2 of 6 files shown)


dunglas:
I would not bother with explicit versions as I'll tag 1.0 soon and this problem will vanish. Other fixes LGTM.

docs/concepts/publishing.md
```diff
@@ -185,6 +185,8 @@ For stricter delivery guarantees (every state change reaches the hub even if the
 
 The Mercure protocol does not require an external hub. An application that already terminates HTTP/2 connections can speak the protocol directly: write SSE bytes to subscribers, validate JWTs, run matchers. This is unusual outside of frameworks that ship their own hub (FrankenPHP, for instance), but the spec allows it.
 
+Check which version of the protocol an embedded hub speaks before you rely on it. [FrankenPHP](https://frankenphp.dev) 1.12.7 embeds `github.com/dunglas/mercure v0.24.2`, so an application running its built-in hub — which includes anything built on the symfony-docker or API Platform distributions — is on 0.x and cannot opt into 1.0 by upgrading the PHP side. Reaching 1.0 there means either running a standalone hub next to the app, or building a custom binary with the 1.0 Caddy module (see [Custom Caddy build](../getting-started/installation.md#custom-caddy-build)).
```
dunglas: Please drop this block. We only document what is in the latest version.

docs/concepts/subscribing.md
```diff
@@ -170,6 +170,26 @@ Fields:
...
+
+```javascript
+// Telling which topic an update belongs to
+es.onmessage = (event) => {
+  const update = JSON.parse(event.data);
+  render(update["@id"], update);
+};
+```
+
+JSON-LD's `@id` is the natural field because it is the topic URL itself, but any identifier works as long as every publisher of that topic sets it. Publishing a bare value with no identifier in it — a number, a status string — leaves a multi-topic subscriber unable to place the update.
```
dunglas: Link to JSON-LD `@id` spec?

docs/concepts/subscribing.md
```diff
@@ -182,21 +202,21 @@ Host: example.com
 200 OK
 Content-Type: application/ld+json
 Link: <https://hub.example.com/.well-known/mercure>; rel="mercure"
+Link: </books/1>; rel="self"
```
dunglas: This is not necessary in this case. The self trick is only useful for Content negotiation (see the spec). Here, using the current URL is good enough to discover the topic.

dunglas (changes requested):

PR #1376, opened 2026-09-07
docs: a Symfony and API Platform page for 1.0

BySplashGm:
Adds `docs/ecosystem/symfony.md`: what it takes to run `symfony/mercure-bundle` 0.5
against a 1.0 hub.

The 1.0 tree has no framework documentation. `ecosystem/awesome`,
`reference/protocol` and the Symfony snippet in `concepts/publishing` link to Symfony
and API Platform with no version requirement and no mention of what 1.0 needs, so a
reader following them lands on 0.x material.

### What the page documents

- **`protocol_version` is per hub and defaults to `0.x`.** Upgrading the bundle
  changes nothing on its own: the app keeps minting legacy `mercure`-claim tokens and
  the 1.0 hub answers `401 invalid_token`.
- **The claims requirement.** Under `protocol_version: '1.0'` with `jwt.secret` or
  `jwt.jwks_uri`, the bundle refuses to compile the container unless `jwt.claims`
  defines `iss`, `sub` **and** `client_id` (RFC 9068). The page quotes the error, so
  it is searchable.
- **`aud` defaults to the hub's `public_url`, then `url`** — wrong in the usual Docker
  split, where the app publishes through an internal URL and the browser uses a public
  one, because a 1.0 hub derives its expected audience from the request. Pin
  `resource_identifier` or set `aud`.
- **`jwt.jwks_uri`** requires `protocol_version: '1.0'` plus `web-token/jwt-library`,
  and is mutually exclusive with `jwt.secret`.
- **`jwt.algorithm` has two spellings**, because the two factories name algorithms
  differently: `hmac.sha256` with `jwt.secret` (`LcobucciFactory`) versus the JWA name
  `HS256` with `jwt.jwks_uri` (`WebTokenFactory`).
- **`cookie_name` flips with `protocol_version`** (`mercureAuthorization` →
  `__Secure-mercure_access_token`), and `__Secure-` needs HTTPS on both sides.
- **The Symfony floor is `^6.4 | ^7.3 | ^8.0`**, not `^6.4 | ^7.0`: an app on
  7.0–7.2 cannot install the bundle at all and has to move to 7.3 first.
- **The subscriber side.** `topic=` → `match=` / `match_urlpattern=`, plus where the
  topic comes from in practice: API Platform publishes on the resource's absolute IRI,
  which is the `@id` of the JSON-LD document it serves, and both `symfony/mercure`'s
  `Discovery` and API Platform advertise `rel="mercure"` only.
- **FrankenPHP's embedded hub is 0.x** (v1.12.7 pins `mercure v0.24.2`), which matters
  because symfony-docker and the API Platform distribution are how most Symfony apps
  run an embedded hub. Includes the build recipe that produces a 1.0-capable binary,
  and notes that not embedding at all is the simpler answer.
- **An upgrade checklist**, in the order the steps actually block each other.

### Placement

`docs/ecosystem/`, next to `awesome` and `conformance-tests`, since that is where the
existing integration material lives — no new nav section. Wired into `docs/README.md`,
`llms.txt`, `ecosystem/awesome`, `reference/protocol` and `concepts/publishing`.

Happy to move it (a `docs/integrations/` section, `getting-started/`, or the bundle's
own README), split the API Platform half out, or cut it down — placement and depth are
your call, and the content stands on its own wherever it goes.

### Verification

Constraints and behavior read from the installed `symfony/mercure-bundle` 0.5
(`composer.json`, `src/DependencyInjection/Configuration.php`,
`src/DependencyInjection/MercureExtension.php`, `CHANGELOG.md`), `symfony/mercure`
0.8 (`src/Discovery.php`), and API Platform 4.3
(`State/MercureLinkProcessor.php`, `Doctrine/EventListener/PublishMercureUpdatesListener.php`).
The configuration shown is the one running in
[`api-platform/esa`](https://github.com/api-platform/esa), verified by publishing to
and subscribing from a `v1.0.0-alpha.3` hub. FrankenPHP's pin was read from `go.mod`
and `caddy/go.mod` at `v1.12.7` (`main` pins the same) and confirmed in the shipped
binary's build info.

Docs only. Relative links and heading anchors across `docs/` were checked to resolve.
`prettier`, `markdownlint` and `textlint` have not been run.


docs/concepts/publishing.md
```diff
@@ -97,6 +97,8 @@ public function __invoke(HubInterface $hub) {
 }
 ```
 
+The code is the same on 0.x and 1.0; what differs is the token the bundle mints for it. Set `protocol_version: '1.0'` on the hub, or it keeps minting 0.x tokens that a 1.0 hub rejects — see [Symfony and API Platform](../ecosystem/symfony.md).
```
dunglas: Let's only document 1.0 at il will be release soon!

dunglas:
Shouldn't put these changes directly in Symfony and API Platform docs instead? This is better to have a single source of truth (clarity and maintenance).

dunglas:
The diff is empty, is it intented?

BySplashGm:
Closing - moving this to symfony/symfony-docs#22973 and api-platform/docs#2323 instead.

PR #1390, opened 2026-09-17
docs: add doc for laravel broadcasting

Lea-Bar:
Inside this PR, I added a documentation about the use case of laravel with mercure.


dunglas:
Thanks!

